PR #1, opened 2022-02-18
Tests linearizability with threads

n-osborne:
Adapt `lin_prop` for domains and threads so that we can reuse `check_seq_cons`.
Tests for linearizability with thread are added in `src/lin_test.ml`

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -11,4 +11,4 @@
 (library
  (name lin)
  (modules lin)
- (libraries qcheck))
+ (libraries threads qcheck))
--- a/lib/lin.ml
+++ b/lib/lin.ml
@@ -16,9 +16,6 @@ module type CmdSpec = sig
   type res
   (** The command result type *)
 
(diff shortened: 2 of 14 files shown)


jmid:
Great - thanks!  👍‍ 😀

A couple of random questions:
1. Have you tried running the tests locally to see if the tests finds thread-unsafe ones (refs, clist, ... perhaps)?
2. I can see why we need to communicate results back to the original thread, but I'm a bit hesitant about the
    new `dummy : res` value, as it adds another definition for a client to specify.
    I would therefore like to understand your reason for choosing this design: 
    Did you observe problems with triggering errors or shrinking when the threads would allocate, e.g., compared to
    building up a result list and storing it in a shared `ref` when a thread is done?
3. The domain tests are repeating each test instance to "make the output more deterministic" 😉
    Did you observe whether the thread tests need less/same/more repetitions to trigger and produce small counterexamples?

PR #2, opened 2022-03-07
Linearizability tests for Queue and Stack module

n-osborne:
- each command is protected by a Mutex
- FIXME: commands that should return an `int` return an `int arbitrary`

### DIFF
--- a/src/stack/lin_tests.ml
+++ b/src/stack/lin_tests.ml
@@ -1,9 +1,8 @@
 open QCheck
 
-module QConf =
+module Spec =
   struct
     type t = int Stack.t
-
     let m = Mutex.create ()
     
     type cmd =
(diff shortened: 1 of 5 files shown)


src/stack/lin_tests.ml
```diff
@@ -0,0 +1,105 @@
...
+      | RTop_opt of int option
+      | RClear
+      | RIs_empty of bool
+      | RLength of int [@@deriving show { with_path = false }]
+
+    let init () = Stack.create ()
+    let cleanup _ = ()
+  end
+
+module QConf =
```
jmid, line 33: `QConf` -> `SConf`  - to reflect that these are stack tests?

src/stack/lin_tests.ml
```diff
@@ -0,0 +1,105 @@
...
+                      with Stack.Empty -> None)
+      | Top_opt  -> RTop_opt (Stack.top_opt q) 
+      | Pop      -> RPop (
+                      try Some (Stack.pop q)
+                      with Stack.Empty -> None)
+      | Pop_opt  -> RPop_opt (Stack.pop_opt q)
+    
+  end
+
+module QMutexConf =
```
jmid, line 52: Similar: `SMutexConf`?

src/stack/lin_tests.ml
```diff
@@ -0,0 +1,105 @@
...
+                    Mutex.unlock m;
+                    RPop r
+      | Pop_opt  -> Mutex.lock m;
+                    let r = Stack.pop_opt q in
+                    Mutex.unlock m;
+                    RPop_opt r
+  end
+
+module QT = Lin.Make(QConf)
+module QMT = Lin.Make(QMutexConf)
```
jmid, line 96: Use `ST` and `SMT`?

jmid:
Looks good!
- I like how you reuse the `Spec`-module by combining it with two different "interpreter"-functions - that's elegant! 
- Minor: I spotted a few identifiers which I commented on inline.
- If you are feeling courageous I've previously experimented with QCheck's ability to generate (pure!) functions, e.g., to test a `fold`-function. Here's an example using it (also to test `Stack`) from `QCSTM` - a predecessor of `STM.ml`: 
  https://github.com/jmid/qcstm/blob/master/examples/stk.ml

PR #5, opened 2022-03-09
Put Lin and STM libraries in a standalone package

OlivierNicole:
This creates packages containing the `Lin` and `STM` libraries, called `lin` and `stm` respectively. It allows users to install and use these libraries in a simple way, independently of the multicoretests test suite—which is restricted to the `multicoretests` package. It also allows to install `lin` and `stm` without some dependencies of multicoretests such as ounit2, lockfree or ppx_deriving.

The package multicoretests now depends on `lin` and `stm`.

### DIFF
--- a/lin.opam
+++ b/lin.opam
@@ -18,8 +18,6 @@ depends: [
 ]
 pin-depends: [
   ["domainslib.0.4.0"          "git+https://github.com/ocaml-multicore/domainslib.git#master"]
-  ["sexplib0.v0.14.0"          "git+https://github.com/patricoferris/sexplib0#5.00"]
-  ["ppxlib.0.25.0~5.00preview" "git+https://github.com/kit-ty-kate/ppxlib#500+sexp"]
 
   ["qcheck-core.0.18.1"        "git+https://github.com/c-cube/qcheck.git#master"]
   ["qcheck-ounit.0.18.1"       "git+https://github.com/c-cube/qcheck.git#master"]
--- a/lin.opam
+++ b/lin.opam
@@ -1,5 +1,6 @@
 opam-version: "2.0"
-synopsis:     "Experimental multicore tests"
+name:         "lin"
+synopsis:     "Experimental library for linearizability checking in Multicore"
 maintainer:   ["Jan Midtgaard <mail@janmidtgaard.dk>"]
 authors:      ["Jan Midtgaard <mail@janmidtgaard.dk>"]
 license:      "BSD-2-clause"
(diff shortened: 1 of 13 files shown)


OlivierNicole:
The readme should probably be adapted to mention the three packages, too.

lin.opam
```diff
@@ -0,0 +1,44 @@
...
+  "ocamlfind"           {>= "1.9.3"}
+  "domainslib"          {>= "0.4.0"}
+  "qcheck-core"         {>= "0.18.1"}
+  "qcheck"              {>= "0.18.1"}
+  "odoc"                {with-doc}
+]
+pin-depends: [
+  ["domainslib.0.4.0"          "git+https://github.com/ocaml-multicore/domainslib.git#master"]
+  ["sexplib0.v0.14.0"          "git+https://github.com/patricoferris/sexplib0#5.00"]
+  ["ppxlib.0.25.0~5.00preview" "git+https://github.com/kit-ty-kate/ppxlib#500+sexp"]
```
jmid, line 22: I think `sexplib0` and `ppxlib` can be removed here.

Pinning recent versions of these two is needed for
- `ppx_deriving` - to auto-derive `show`
- `ppx_deriving_qcheck` - to auto-derive generators
if I remember correctly.

A user should be able to use the library with hand-written versions of the above, but both are just very convenient.
I guess that information would be good to include in the README...

stm.opam
```diff
@@ -0,0 +1,44 @@
...
+  "ocamlfind"           {>= "1.9.3"}
+  "domainslib"          {>= "0.4.0"}
+  "qcheck-core"         {>= "0.18.1"}
+  "qcheck"              {>= "0.18.1"}
+  "odoc"                {with-doc}
+]
+pin-depends: [
+  ["domainslib.0.4.0"          "git+https://github.com/ocaml-multicore/domainslib.git#master"]
+  ["sexplib0.v0.14.0"          "git+https://github.com/patricoferris/sexplib0#5.00"]
+  ["ppxlib.0.25.0~5.00preview" "git+https://github.com/kit-ty-kate/ppxlib#500+sexp"]
```
jmid, line 22: I think the same applies here.

jmid:
Thanks a bunch for this!
I think there's a couple of unused pin-depends, which I commented on inline.

You are very welcome to add a short mention of the three opam packages to the README
now that I've merged the other README PR.

Having to make package-annotations on each rule saddens me.
Is there by chance a way to mark an entire (sub-)directory part of the `multicoretests` package?

lin.opam
OlivierNicole: Adressed.

OlivierNicole:
> Having to make package-annotations on each rule saddens me.
> Is there by chance a way to mark an entire (sub-)directory part of the `multicoretests` package?

I'm only starting with dune and reading the docs, but maybe using the [scopes feature](https://dune.readthedocs.io/en/stable/concepts.html#scopes-1)? IIUC this implies to put all the tests and the `multicoretests.opam` file in a subtree.

OlivierNicole:
To avoid the many `(package multicoretests)` annotations, I have tried to use dune scopes in two ways:

- Put all the tests in a `multicoretests/` subdirectory, along with `multicoretests.opam`.
- Put `lin.opam` and `stm.opam`.

Both failed with errors like:
```
Entering directory '/home/olivier/opam-pins/multicoretests'
File "lib/dune", line 8, characters 14-17:
8 |  (public_name stm)
                  ^^^
Error: The current scope doesn't define package "stm".
The only packages for which you can declare elements to be installed in this
directory are:
- multicoretests (because of multicoretests.opam)
```
That is, dune seems to ignore the `<package>.opam` files when they are not at the project root. I am here on unfamiliar grounds, so to go further I would need to ask advice to more knowledgeable users.

jmid:
OK, thanks for trying!
I'll merge and then we can adjust later (if we figure it out).

Issue #8, opened 2022-03-09
Spurious ocamlbuild pin?


OlivierNicole:
The current opam spec for the multicoretests packages [pins ocamlbuild](https://github.com/jmid/multicoretests/blob/main/multicoretests.opam#L33), which surprises me. Is it still relevant?

jmid:
I think you are right. 
`ocamlbuild` 0.14.1 has been released: https://github.com/ocaml/opam-repository/pull/20697
and seems to support 5.00: https://github.com/ocaml/ocamlbuild/releases/tag/0.14.1

Looking at https://github.com/ocaml/ocamlbuild/releases - 0.14 was released 3 years ago(!) 
and there has been a few fixes since then, so it makes sense to have gone with trunk.
Actually, I think I used trunk as a "multicore installable version" to install kcas and lockfree
which both listed ocamlbuild as a dependency at some point.

A month ago I realized that lockfree didn't have a lot of dependencies anymore - and submitted
a PR to clean them up: https://github.com/ocaml-multicore/lockfree/pull/10
Digging a bit more, it seems both `ocamlbuild` and `ocamlfind` are not actual dependencies for kcas either:
 https://github.com/ocaml-multicore/kcas/pull/4
I've just submitted a PR to remove them: https://github.com/ocaml-multicore/kcas/pull/12

Once that PR is merged we should be able to simply remove the ocamlbuild pin.
Here-and-now, using my patched fork should also work:
```
  ["kcas.0.14"                 "git+https://github.com/jmid/kcas#clean-up-deps"]
```

jmid:
This is fixed by #12

PR #12, opened 2022-03-10
Remove dependency to ocamlbuild

OlivierNicole:
Following the discussion in #8. The dependencies have been updated upstream, so it seems safe to remove our ocamlbuild pin.


jmid:
Thanks! :-)

PR #13, opened 2022-03-10
Add (failing) {In,Out}_channel linearization tests

OlivierNicole:
Note: this depends on #5, which is why I open it as a draft PR.

This adds linearizability testing for a few functions of the new Stdlib modules `In_channel` and `Out_channel`. As of now these functions are not thread-safe, so the tests fails, sometimes with segfaults (due to concurrent `close` and output functions in `Out_channel`).

Left to do:
- [ ] Update dune file appropriately once the fate of #5 is settled.
- [ ] Determine how this (currently failing) can be integrated properly into the test suite.
- [ ] Maybe enrich the test to cover more functions once their precise contract is decided by the compiler team (fix, or not fix and document, etc.)

### DIFF
--- a/src/in_out_channel/dune
+++ b/src/in_out_channel/dune
@@ -1,0 +1,29 @@
+;; Linearizability tests of the stdlib Queue library
+
+;; this prevents the tests from running on a default build
+(alias
+ (name default)
+ (package multicoretests)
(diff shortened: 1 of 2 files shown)


OlivierNicole:
Closing as this is now superseeded by #270.

PR #17, opened 2022-03-11
Fix build instructions in readme

OlivierNicole:
I realized that my own instructions were wrong. Simply running `opam install . --deps-only` and `opam exec -- dune build` leads to the error that packages `lin` and `stm` are not found. One must first install them.

Note that this is due to lin and stm not being part of a public repository. Once they are, we can remove this step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,19 +78,29 @@ package](lin.opam).
 Installation instructions, and running the tests
 ================================================
 
-The test suite can be run with the following commands:
-```
-opam install . --deps-only --with-test
-opam exec -- dune build
-opam exec -- dune runtest
-```
-
-Package `lin` and `stm` can be installed with:
+Package `lin` and `stm` can be installed independently with:
 ```
 opam install ./lin.opam
 opam install ./stm.opam
 ```
 
+The test suite can be run with the following commands (from the root of this directory):
+```
+opam install . --deps-only --with-test
+dune build
+dune runtest -j1 --error-reporting=deterministic
+```
+
+Individual tests can be run by invoking `dune exec`. For example:
+```
+$ dune exec src/atomic_test.exe -- -v
+random seed: 51501376
+generated error fail pass / total     time test name
+[✓] 1000    0    0 1000 / 1000     0.2s sequential atomic test
+[✓] 1000    0    0 1000 / 1000    21.7s parallel atomic test (w/repeat)
+================================================================================
+success (ran 2 tests)
+```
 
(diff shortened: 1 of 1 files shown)


jmid:
> I realized that my own instructions were wrong. Simply running opam install . --deps-only and opam exec -- dune build leads to the error that packages lin and stm are not found. One must first install them.

Are you sure?
I just did a fresh clone and double-checked the output of `opam list` and `opam pin list` to make sure that I don't have `stm` and `lin` installed or pinned. Here the 3 listed commands run without a missing package error:
```bash
$ opam install . --deps-only --with-test
$ opam exec -- dune build
$ opam exec -- dune runtest
```

1. I think we can drop the `opam exec --` prefix from the last two commands though
2. I believe that using `dune runtest -j1 --error-reporting=deterministic` paves for better, more deterministic errors and error messages
3. We may want to mention that one can run individual tests by invoking
    ```
      $ dune exec src/atomic_test.exe -- -v
      random seed: 51501376
      [✓] 1000    0    0 1000 / 1000     0.2s sequential atomic test
      [✓] 1000    0    0 1000 / 1000    21.7s parallel atomic test (w/repeat)
      success (ran 2 tests)
    ```

jmid:
> I realized that my own instructions were wrong. Simply running opam install . --deps-only and opam exec -- dune build leads to the error that packages lin and stm are not found. One must first install them.

Are you sure?
I just did a fresh clone and double-checked the output of `opam list` and `opam pin list` to make sure that I don't have `stm` and `lin` installed or pinned. Here the 3 listed commands run without a missing package error:
```bash
$ opam install . --deps-only --with-test
$ opam exec -- dune build
$ opam exec -- dune runtest
```

1. I think we can drop the `opam exec --` prefix from the last two commands though
2. I believe that using `dune runtest -j1 --error-reporting=deterministic` paves for better, more deterministic errors and error messages
3. We may want to mention that one can run individual tests by invoking
    ```
      $ dune exec src/atomic_test.exe -- -v
      random seed: 51501376
      [✓] 1000    0    0 1000 / 1000     0.2s sequential atomic test
      [✓] 1000    0    0 1000 / 1000    21.7s parallel atomic test (w/repeat)
      success (ran 2 tests)
    ```

OlivierNicole:
Mmh, all right, I did get a `library STM not found` error but I can't seem to reproduce it. It must have been caused by something else.

Your three suggestions look good to me, I can add them to the patch.

OlivierNicole:
Done.

PR #19, opened 2022-03-15
Put lin and stm in a single multicorecheck package

OlivierNicole:
The current package organization makes it difficult for `Lin` and `STM` to have modules in common without introducing a third package. Rather than having a user-facing package without a clear purpose, this puts Lin and STM in a single `multicorecheck` package. Common dependencies can exist in the form of private library in the same package (akin to `lib/util`).


jmid:
Looks good - thanks!

We should remember to update the README,md to reflect the new structure.

jmid:
BTW, I think we can remove the dependency on `ocamlfind`:
  https://github.com/jmid/multicoretests/blob/77b370a15f18c35a6dbe28a669aacd3a6b9a85ac/multicorecheck.opam#L13

OlivierNicole:
Done in the new version.

OlivierNicole:
Restructured to have all three libraries back into the same directory.

PR #23, opened 2022-03-21
Avoid double cleanup in Lin

### LINKED ISSUE #9: Unexpected double call of `cleanup` by Lin

jmid:
This fixes #9.

The search for a linearization may explore multiple sequential runs, which each require an `init` and a `cleanup`.
The exploration may stop early and try another combination, which means we cannot simply `cleanup` afterwards.
Instead we call `cleanup` when the exploration of a linearization is done.

### DIFF
--- a/lib/lin.ml
+++ b/lib/lin.ml
@@ -84,29 +84,31 @@ module Make(Spec : CmdSpec) (*: StmTest *)
 
   let rec check_seq_cons pref cs1 cs2 seq_sut seq_trace = match pref with
     | (c,res)::pref' ->
-       res = Spec.run c seq_sut
-       && check_seq_cons pref' cs1 cs2 seq_sut (c::seq_trace)
+        if res = Spec.run c seq_sut
+        then check_seq_cons pref' cs1 cs2 seq_sut (c::seq_trace)
+        else (Spec.cleanup seq_sut; false)
+    (* Invariant: call Spec.cleanup immediately after mismatch  *)
     | [] -> match cs1,cs2 with
-            | [],[] -> true
+            | [],[] -> Spec.cleanup seq_sut; true
             | [],(c2,res2)::cs2' ->
-               res2 = Spec.run c2 seq_sut
-               && check_seq_cons pref cs1 cs2' seq_sut (c2::seq_trace)
+                if res2 = Spec.run c2 seq_sut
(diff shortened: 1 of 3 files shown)


jmid:
@OlivierNicole This turned out to be a little more subtle than I had anticipated.
Could you give this a spin and/or a look?

OlivierNicole:
The changes look good to me. I confirm that it removes the exception I complained about in #9.

PR #27, opened 2022-03-30
add test for stdlib ephemeron module

n-osborne:


src/ephemeron/dune
```diff
@@ -0,0 +1,28 @@
...
+ (preprocess (pps ppx_deriving_qcheck ppx_deriving.show)))
+
+(rule
+ (alias runtest)
+ (action
+  (progn
+   (with-accepted-exit-codes 1
+    (with-stdout-to "lin-output.txt" (run ./lin_tests.exe --no-colors --verbose)))
+   (cat lin-output.txt)
+    (run %{bin:check_error_count} "lin_tests" 1 lin-output.txt))))
```
jmid, line 28: Why do you expect 1 error precisely?
Are ephemeron's not domain-safe - or is it failing for another reason?

src/ephemeron/lin_tests.ml
```diff
@@ -0,0 +1,61 @@
...
+      | Add of int' * string
+      | Remove of int'
+      | Find of int'
+      | Find_opt of int'
+      | Find_all of int'
+      | Replace of int' * string
+      | Mem of int'
+      | Length
+      | Clean [@@deriving qcheck, show { with_path = false }]
+    and int' = int64 [@gen fun st -> Gen.nat st |> Int64.of_int]
```
jmid, line 20: It may be a good idea to test with a boxed type like `int64`.
Can you add a comment why you chose to do so?

jmid (commented):
Great - thanks! This is very clean code - beautiful!
I've commented on a few things that puzzled me.

src/ephemeron/dune
n-osborne: Ephemeron are failing the test with domains. But it shouldn't.
Expected behaviour is to have `0` here. I'll correct it.

jmid:
This looks great!

I believe the CI is failing for an unrelated reason.
This looks like a ppxlib version incompatibility and may be caused by the OCaml 5.0 happening yesterday....

src/ephemeron/dune
```diff
@@ -0,0 +1,41 @@
+;; Linearizability tests of the stdlib Ephemeron module
+
+;; this prevents the tests from running on a default build
+(alias
+ (name default)
+ (deps lin_tests_dsl.exe))
```
jmid, line 6: `stm_test.exe` is missing here as a dependency on "`default`".
This ensures that it is compiled when `dune build` is called.

jmid: Done in f277d63

jmid:
Commit f29ca26 attempts to relax the postcondition to express that 'missing expected entry' (because of GC) is also OK.
With that, I am get counterexamples closer to those from `Hashtbl` (both modules are not concurrency safe https://github.com/ocaml/ocaml/pull/11193).
Sometimes I get a weird and longer counterexample though... :grimacing:

PR #33, opened 2022-04-12
Use customizable equality in Lin

OlivierNicole:
Polymorphic equality is sometimes not what you want. I encountered the problem when I needed to instantiate type `res` to values containing a type witness, an `'a -> 'a -> bool` equality function and a value of type `'a`.

This proposes to add an `equal_res` function to the specification given to `Lin` and updates the tests accordingly.


jmid:
This looks very sensible! I'll just let the CI finish before merging...

OlivierNicole:
Oops, it looks like `dune build` does not build all the tests. Fixing…

PR #42, opened 2022-04-14
Statistics script

jmid:
This PR contains a statistics script handy for deciding whether changes influence our ability to detect issues


OlivierNicole:
That's very clear, thank you!

PR #49, opened 2022-04-21
Switch to branch with improved shrinkers

jmid:
This PR pins the latest QCheck shrinker improvements from c-cube/qcheck#242.
I've rebased this branch a few times and let it run on the CI without any issues.

Overall, because of the reduced number of shrink attempts I expect it to result it potentially smaller counterexamples.
If, say `repeat` and `retry` bring repeatability of an issue to 95% 
- a shrink sequence of 5 steps will succeed with 76% chance whereas
- a shrink sequence of 10 steps will succeed with 57% chance
so the fewer required steps, the better.

### DIFF
--- a/multicorecheck.opam
+++ b/multicorecheck.opam
@@ -15,9 +15,9 @@ depends: [
   "odoc"                {with-doc}
 ]
 pin-depends: [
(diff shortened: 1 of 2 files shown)


jmid:
Closing as c-cube/qcheck#242 has now been merged

PR #50, opened 2022-04-22
Update to lockfree 0.2.0

### LINKED ISSUE #35: Update lockfree tests to 0.2
Update the tests to target the rewamped 0.2 release: https://github.com/ocaml-multicore/lockfree

This involves
- a name change - from `master` to `main` as the default
- moving the `Domainlib.Ws_deque` test

jmid:
This fixes #35.

- The PR moves the domainslib/ws_deque_test to lockfree - which for now only contains this one data structure.
- The old lock tests are therefore removed.
- Finally it adapts the ws_deque test to test both positively and negatively


jmid:
I checked locally that this runs without issues on MacOSX on both ocaml trunk commit 7585460 and ocaml PR#11214.
I will therefore merge this and expect the CI to fail until PR#11214 or a similar fix is merged in ocaml trunk.

jmid:
(the Linux failure happens in the `Effect` interpretation where an expected error is not found - and is thus unrelated to the lockfree and domainslib changes)

Issue #40, opened 2022-04-13
Port examples from qcstm


jmid:
STM's predecessor https://github.com/jmid/qcstm
contains a range of examples https://github.com/jmid/qcstm/tree/master/examples

We should port these to STM which would allow to run them both sequentially and in parallel mode.

The primary difference is that `qcstm`'s `run_cmd : cmd -> state -> sut -> bool` has two tasks which is broken in two in `STM`:
```ocaml
  val run : cmd -> sut -> res                 (* for interpreting the command over the system under test *)
  val postcond : cmd -> state -> res -> bool  (* for checking the post condition *)
```

Issue #38, opened 2022-04-12
Fix precond in STM's parallel mode


jmid:
STM should be fixed to respect preconditions `precond` in parallel mode - `agree_test_par`.
Currently this is only the case in sequential mode - `agree_test`.

jmid:
Closing as this has been fixed with the initial generate-and-filter implementation in #67

PR #58, opened 2022-04-27
port buffer example from QCSTM

n-osborne:
This PR ports the `Buffer` example from QCSTM to STM (issue #40)
allowing to check that Buffer is **not** Domain safe.

The QCSTM Buffer example can be found [here](https://github.com/jmid/qcstm/blob/master/examples/buf.ml)

- `next_state` has been slightly changed
- a `type res` has been added
- `run_cmd` has changed into `run` and `postcondition` has been added

There are two tests that are run:

- sequential test that is expected to succeed
- parallel test that is expected to fail

### DIFF
--- a/src/buffer/buffer_stm_test.ml
+++ b/src/buffer/buffer_stm_test.ml
@@ -1,0 +1,123 @@
+open QCheck
+open Util
+
+(** parallel STM tests of Buffer *)
+
+(* port from the QCSTM example *)
+module BConf =
+struct
+  type cmd =
+    | Contents
+    (* To_bytes | Sub | Blit *)
+    | Nth of int
+    | Length
+    | Clear
+    | Reset
+    | Add_char of char
+    (* Add_utf8_uchar | Add_utf_16le_uchar | Add_utf_16be_uchar *)
+    | Add_string of string
+    (*| Add_bytes of bytes*)
+    | Truncate of int
+  [@@deriving show { with_path = false }]
+
+  type state = char list (* in reverse *)
+  type sut = Buffer.t
+
+  let arb_cmd s =
+    QCheck.make ~print:show_cmd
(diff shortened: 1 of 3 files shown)


src/buffer/buffer_stm_test.ml
```diff
@@ -0,0 +1,123 @@
...
+    | Contents
+    (* To_bytes | Sub | Blit *)
+    | Nth of int
+    | Length
+    | Clear
+    | Reset
+    | Add_char of char
+    (* Add_utf8_uchar | Add_utf_16le_uchar | Add_utf_16be_uchar *)
+    | Add_string of string
+    (*| Add_bytes of bytes*)
```
jmid, line 19: I know this is a port from `qcstm` - so this is not to point any fingers.
Still: how hard would it be to include the remaining bindings from `Buffer`'s signature?

src/buffer/buffer_stm_test.ml
```diff
@@ -0,0 +1,123 @@
...
+        | [],_ -> []
+        | _::_,0 -> []
+        | c::cs,_ -> c::trunc cs (n-1) in
+      List.rev (trunc (List.rev s) i)
+  
+  let init_sut () = Buffer.create 16
+  let cleanup b   = Buffer.reset b
+
+  let precond c s = match c with
+    | Truncate i -> i >= 0 && i <= List.length s
```
jmid, line 71: Just a word of warning: `precond` is ignored in parallel mode currently #38.
I hope to fix it shortly though, so it is fine to just keep the code as above.

jmid (changes requested):
Looks good - thanks!
I had a few comment - and found one `postcond` simplification (I think).

jmid:
Looks good - thanks!

PR #63, opened 2022-05-10
STM clean-up

jmid:
This PR cleans up the STM signature, leaving the `repeat`-`retries` combination property, which has worked well for Lin.

Currently, the numbers are set to `repeat:25` and `retries:15`. We may want to adjust these before merging.

As a bonus
- it adds a negative `int64 ref` test
- removes a bit of whitespace

As an extra bonus, with the changes the `Buffer` test can segfault my local Linux running:
```
$ dune exec src/buffer/buffer_stm_test.exe -- -v -s 81407870
```
I'm therefore curious how the Linux CI reacts...


jmid:
I should add that the `Lazy` tests take a while to run with these parameter numbers, which is why I am curious to reduce them...

jmid:
Rerunning jobs as trunk has now been patched for the MacOS segfault.

PR #67, opened 2022-05-11
Initial STM precond fix, add Domainslib.Chan tests

jmid:
This PR provides an initial implementation STM `precond` support for parallel mode using a generate-and-filter approach.
Following Claessen-al:ICFP09 we keep only `cmd list` triples where all interleavings satisfy all preconditions.

The `precond` implementation is enough to get the second part running:
This part tests the `Domainslib.Chan` module. This requires `precond` to avoid
- `recv` blocking on an empty channel
- `send` blocking on a full channel

For the `Domainslib.Chan` tests, this means that we need to generate around twice as many triples as the test `count`.
I've also checked that the run time of the other STM tests is not affected noticably. For reference we now have the following STM tests:
```
src/atomic/atomic_test.ml
src/buffer/buffer_stm_test.ml           ;; expected to fail
src/domainslib/chan_tests.ml
src/lazy/lazy_stm_test.ml               ;; expected to fail
src/lockfree/ws_deque_test.ml
src/neg_tests/conclist_test.ml          ;; expected to fail
src/neg_tests/ref_test.ml               ;; expected to fail
```

Ping @n-osborne :eyes:

### DIFF
--- a/lib/STM.ml
+++ b/lib/STM.ml
@@ -267,7 +267,7 @@ struct
   let agree_test_par ~count ~name =
     let rep_count = 25 in
     let seq_len,par_len = 20,12 in
-    let max_gen = 2*count in (* precond filtering may require extra generation: max. 2*count though *)
+    let max_gen = 3*count in (* precond filtering may require extra generation: max. 3*count though *)
     Test.make ~retries:15 ~max_gen ~count ~name:("parallel " ^ name)
       (arb_cmds_par seq_len par_len)
       (repeat rep_count agree_prop_par) (* 25 times each, then 25 * 15 times when shrinking *)
--- a/lib/STM.ml
+++ b/lib/STM.ml
@@ -167,7 +167,35 @@ struct
     let res_arr = Array.map (fun c -> Domain.cpu_relax(); Spec.run c sut) cs_arr in
     List.combine cs (Array.to_list res_arr)
 
-  let rec check_obs pref cs1 cs2 s = match pref with
+  (* checks that all interleavings of a cmd triple satisfies all preconditions *)
+  let rec all_interleavings_ok pref cs1 cs2 s =
+    match pref with
+    | c::pref' ->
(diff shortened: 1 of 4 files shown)


lib/STM.ml
```diff
@@ -171,7 +171,35 @@ struct
     let res_arr = Array.map (fun c -> Domain.cpu_relax(); Spec.run c sut) cs_arr in
     List.combine cs (Array.to_list res_arr)
 
-  let rec check_obs pref cs1 cs2 s = match pref with
+  (* checks that all interleavings of a cmd triple satisfies all preconditions *)
+  let rec all_interleavings_ok pref cs1 cs2 s =
+    match pref with
```
n-osborne, line 176: As the prefix is sequential, it should be valid by construction (?). So, do we really need to check it here or could we just compute the `spawn_state`?

jmid: Good point! :+1: 
From a generation perspective I completely agree.
From a shrinking perspective, I don't think we can be sure that the prefix continues to be valid `precond`-wise, e.g., if the `cmd list` shrinker removes `cmd`s from the sequential prefix...

Such shrinking reminds me that it would be good to actually test that this combination works - a negative test (triggering shrinking) with a non-constant `precond`....

jmid:
Rerunning jobs as trunk has now been patched for the MacOS segfault.

jmid:
Rebased on top of #63 - also bumping `max_gen` to `3 * count`...

jmid:
Ran this manually on the MacMini also - 5 times.
- 3 runs were all green
- one run triggered the `Buffer` failure (first time on MacOS I think)
- another run failed to trigger the expected 2 failures in `src/lazy/lazy_stm_test.ml`

The latter only runs 100 iterations though - so bumping this `count` to 200...

jmid:
Runs fine on manual MacOS tests - merging

PR #68, opened 2022-05-17
Try MacOS 10.15 and 12 too

jmid:
This is just a test to see if the MacOS segfault is triggered on other versions


jmid:
Closing as switching to the opam alpha repo did the trick

PR #69, opened 2022-05-17
Wrap domain functions in a handler to ensure proper joining

jmid:
When the first domain function raises an exception, we risk not joining the second domain. This is bad for reproducability. 
Furthermore, unjoined underlying pthreads at exit can apparently cause crashes on Windows (not good either...).
We fix this by wrapping the function bodies in an exception handler and reraising an exception in the "root domain".

Thanks to @patricoferris for pointing this out :pray:


jmid:
Rebased on main, hoping for more green lights...

PR #70, opened 2022-05-17
Omit buffer test

jmid:
The `Buffer` test can trigger a segfault as described in #63.
This KR therefore
- omits the test (temporarily) and
- wraps the domains functions in exception handlers (building on #69).

The hypothesis is that collectively the two will restore green CI lights for MacOS... :crossed_fingers:

### DIFF
--- a/lib/STM.ml
+++ b/lib/STM.ml
@@ -219,23 +219,24 @@ struct
     make ~print:(print_triple_vertical Spec.show_cmd) ~shrink:shrink_triple gen_triple
 
   (* Parallel agreement property based on [Domain] *)
-  let agree_prop_par =
-    (fun (seq_pref,cmds1,cmds2) ->
-      assume (cmds_ok Spec.init_state (seq_pref@cmds1));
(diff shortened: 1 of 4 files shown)


jmid:
This is still failing with a segfault https://github.com/jmid/multicoretests/runs/6479465240?check_suite_focus=true#step:6:916

New hypothesis:
- this is caused by `src/lazy_lin_test.ml`
- it is tied to raising exceptions (here `CamlinternalLazy.Undefined`)
- it happens when finishing the test (completing the run of the first of 3 variants - so not at exit)

Still not reproducible locally - so trying to add missing handlers (while crossing fingers...)

jmid:
Hypothesis shot down: last two MacOS CI runs resulted in 4 and 2 segfaults, respectively:
- https://github.com/jmid/multicoretests/runs/6483633874?check_suite_focus=true#step:6:501
- https://github.com/jmid/multicoretests/runs/6483633586?check_suite_focus=true#step:6:441

PR #72, opened 2022-05-21
Initial Hashtbl port to DSL

jmid:
This is an initial port of the `Hashtbl` example to the signature DSL. 
I used this as an illustration at the Tarides presentation on Friday.
It adds `nat_small` to the DSL as part of the PR.


jmid:
Rebased on latest main to avoid noise from MacOS failure

PR #74, opened 2022-05-30
Reduce pin-depends

jmid:
This PR removes two unneeded pin-depends entries.
Both `sexplib0` and `ppxlib` are now available in the alpha repository.


jmid:
Rebased this PR as things are starting to turn more green lately...

Issue #15, opened 2022-03-11
Switch to qcheck-alcotest?


jmid:
The native QCheck_runner is nice for running tests in the terminal, however for the CI less so:
- The non-`verbose`-mode is too silent and hence doesn't give the user a sense of progress (e.g., are we hitting an infinite loop?)
- The responsive UI in `verbose`-mode will overflow the CI-logs with too many redundant lines

For these reasons, I initially just reduced the timing between print-outs:
  https://github.com/jmid/multicoretests/blob/6f086982931c19a970c14deb4d783b42a37bc4b9/lib/util.ml#L9-L13
Unfortunately one size (or timing) does not fit all:
- For quick tests, this will only print a few lines: https://github.com/jmid/multicoretests/runs/5505103534?check_suite_focus=true#step:6:76
- For slower test, this will still print many redundant/noisy lines: https://github.com/jmid/multicoretests/runs/5505103534?check_suite_focus=true#step:6:143

Later, `check_error_count` was added to check ensure that negative tests would fail as expected.
By saving output to a temporary file (for later counting), the "user's sense of progress" was however reduced again.

The current CI logs are thus not in a good state.
We should consider switching to the qcheck-alcotest backend, as alcotest seems to offer
- support for negative tests
- a reasonable UI suitable for CI-logs

Trying it out first on a few smaller examples is probably a good idea to confirm the above.

jmid:
I took an initial stab at this, trying it out on `src/neg_tests/domain_lin_tests.ml`
The experiments are available here: https://github.com/jmid/multicoretests/tree/try-alcotest

Alcotest offers
 - a compact/non-compact switch
 - a less noisy verbose printer
 - continuous output as tests are run
 - marking tests as positive/negative with the above

Assuming one wants
 - negative tests with shrunk output printed, the above output is a bit hard to tell apart:
   ```
     ...     [test family]        0   Test name
   test `Test name` failed on ≥ 1 cases:
   [...]
     [OK]    [test family]        0   Test name
   ```

Seeing the found counterexamples in the CI-log can be useful to tell if shrinkers are working as expected.


Alcotest is missing
 - timings on individual tests - which can be nice to tell what is taking the time

jmid:
With #140 extending the interval between printed CI log messages, this is no longer as relevant.

PR #78, opened 2022-06-01
Utilize negative tests in QCheck

### LINKED ISSUE #29: Mark tests as negative in test source code
For now we are running both positive tests ("things work as expected") and negative tests ("things fail as expected") through QCheck.
This is handled by counting failures with a somewhat clunky `check_error_count`.

This approach has several down sides:
- spraying `dune` files with magic constants as arguments to `check_error_count`
- decoupling the expected behavior from the test source code
- making it harder to pin-point a failure ("we expected 3 errors but got only 2...").

Ideally, this approach should be replaced by being able to mark PBTs as negative: `Test.make_negative ...`
so that the test framework can pin-point and name the test if a negative one should unexpectedly succeed.

jmid:
This is a WIP PR to try out the negative testing primitive `Test.make_neg` from c-cube/qcheck#244.
It is intended to fix #29

Overall
- this allows negative tests to be marked as such with `neg_agree_test_par` directly in the source code
- the clunky `check_error_count` can be eliminated entirely
- magical constants denoting `1`, `2`, and `4` errors in `dune` files are eliminated, and
- `dune` files are generally simplified

As an added bonus, it opens up for switching to an alcotest runner as suggested in #15.

I would therefore argue that this is a clear win!  \o/


jmid:
Rebased to latest `main`

jmid:
`Test.make_neg` is now merged into QCheck master, so this can be picked up again.

jmid:
Rebased after merging #90

jmid:
Rebased on latest main, having QCheck 0.19 with `Test.make_neg` as a dependency.

jmid:
I just annotated the new DSL-ported tests with suitable `neg_` suffixes.
Planning to merge this assuming the CI doesn't complain too much.

jmid:
(Hopefully) last rebasing on `main`

PR #80, opened 2022-06-02
First attempt to reduce boilerplate with type res

OlivierNicole:
This is a first attempt to reduce the boilerplate associated with type `res` when defining STM tests. To see what would change concretely for the STM user, one must look at file `src/hashtbl/stm_test.ml`.

For the sake of the example, the possibility of returning values of type `exn` is not implemented in STM, and the test writer had to extend it. This is done by adding a constructor to the extensible type `'a STM.Res.ty`, and specifying how results of this type should be compared and printed.

A value of type `res` is constructed as `Res (spec, v)` where `spec` is a “type specification” which identifies the type of the result and contains the comparison and printing functions, and `v` is a value. For convenience, STM provides combinators to construct type specifications.

Finally, in `postcond`, instead of matching over the constructors of `res`, the test writer has to match over the possible result types.

### DIFF
--- a/src/hashtbl/stm_test.ml
+++ b/src/hashtbl/stm_test.ml
@@ -1,5 +1,5 @@
 open QCheck
-open Util
+open STM
 
 (** parallel STM tests of Hashtbl *)
 
@@ -79,7 +79,6 @@ struct
     | Length        -> s
 
   let run c h =
-    let open STM.Res in
     match c with
     | Clear         -> Res (unit, Hashtbl.clear h)
     | Add (k,v)     -> Res (unit, Hashtbl.add h k v)
@@ -95,7 +94,6 @@ struct
 
   let precond _ _ = true
   let postcond c (s : state) res =
-    let open STM.Res in
     match c,res with
     | Clear,         Res ((Unit,_),_)
     | Add (_,_),     Res ((Unit,_),_)
--- a/src/hashtbl/stm_test.ml
+++ b/src/hashtbl/stm_test.ml
@@ -45,10 +45,6 @@ struct
     | Mem of char
     | Length [@@deriving show { with_path = false }]
 
-  type _ STM.Res.ty += Exn : exn STM.Res.ty
-
-  let exn : exn STM.Res.ty_show = (Exn, Printexc.to_string)
-
   let init_sut () = Hashtbl.create ~random:false 42
   let cleanup _ = ()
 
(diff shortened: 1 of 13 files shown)


src/hashtbl/stm_test.ml
```diff
@@ -45,16 +45,10 @@ struct
...
-    | RFind of (int, exn) result
-    | RFind_opt of int option
-    | RFind_all of int list
-    | RReplace
-    | RMem of bool
-    | RLength of int [@@deriving show { with_path = false }, eq]
+  type _ STM.Res.ty += Exn : exn STM.Res.ty
+
+  let exn : exn STM.Res.show_eq =
+    { ty = Exn; show = Printexc.to_string; eq = (=) }
```
OlivierNicole, line 51: Type `res` is no longer declared in the specification, but possibly extended.

OlivierNicole:
I was able to remove the clobbering type annotations that I had introduced in the implementations of `postcond`. The typechecker was complaining on lines like:
```ocaml
 let postcond c s res =
    let open STM.Res in
    match c,res with
    | Get,   Res ((Int,_),v)  -> v = s (* error here *)
    | Set _, Res ((Unit,_),_) -> true
    | Add _, Res ((Unit,_),_) -> true
    | Incr,  Res ((Unit,_),_) -> true
```
because `v` has a locally abstract type coming from the existentially-quantified type in `Res` and `s` has type `'a`, and unifying them would cause the type of `v` to appear in the type of the function and thus escape its scope. It is enough to specify that `s` is of type `state` to get rid of such errors.

@jmid I see that you used functions like `Int.equal` instead of the polymorphic `(=)`, which gets rid of the error too. I left them as-is and just removed the single explicit type annotation that you had to write.

jmid:
This is great!
I indeed discovered the `Int.equal` solution to solve exactly this.
I think `s : state` and reverting to `(=)` may read nicer though - so feel free to revert my change!

OlivierNicole:
Done.

jmid:
Having ported a number of examples, I already tend to forget `open`ing the `STM.Res` sub-module.
There isn't a lot in `STM` though: We mostly use `STM.Make` and `STM.protect` (from `include `Util`).
How much would break if we unite the two name spaces (or `include STM.Res` in `STM`)?

I'm thinking we want something as simple as possible to use.
If an `open STM` at the top is sufficient - it should make for an even simpler user experience... :thinking:

OlivierNicole:
I just made the change; the result looks OK enough IMO.

jmid:
Looks good indeed - thanks! :smiley: 

I think this is ready to merge (I'll let the CI finish first though). Do you agree?

OlivierNicole:
All seems good to me, yes!

Issue #66, opened 2022-05-11
Split Lin interpretations


jmid:
The MacOS segfault that KC just fixed in ocaml/ocaml#11250 reveals that the `threads` module performs a number of side-effects when loaded. By having all three of our Lin interpretations (Domain,Thread,Fiber) in the same module requiring `threads` (even though it isn't used) this means
- we confine our test runs to those where `thread` is always loaded and its callbacks are registered - see https://github.com/ocaml/ocaml/blob/37dec39a627b3b8abcf84c2463c64d2336b2bee6/otherlibs/systhreads/thread.ml#L87-L98
- Domain tests will run slower than necessary because of the overhead

The latter is naturally less serious than the first.

We should consider splitting the three interpretations up into separate files with separate dependencies such that only the Lin `Thread` interpretation requires `threads`.

jmid:
Ideally a design would
- allow us to keep a Lin (or STM) specification in one, common file
- run both a Domain or Thread interpretation from the same specification (for either Lin or STM)
- allow only the `Thread` interpretation to require `threads`
- (be as simple as possible)

This way
- a specification is reusable
- a pure Domain interpretation executable does not perform the side-effecting load of `threads`
- one is still able to test this corner-case configuration by loading `threads` nevertheless
- one can write the "alternative printer interpretation" (`Domain-safety:  Atomic: OK ... Hashtbl: No ...`) we initially talked about

n-osborne:
I've begun to experiment with building the `Domain` and the `Thread` tests with two different functors in two different files (so we have a `STM_Domain.Make(Spec)` and a `STM_Thread.Make(Spec)`), most of the work being done by an intermediary common functor.

There is the question of the sequential `agree_test`. Do we want also a `STM_Sequential.Make(Spec)`?
I think it would be cleaner, but maybe there are other options I didn't consider?

jmid:
Good point! :+1: 
I think a separate sequential functor would be good - akin to what `qcstm` offers.

jmid:
Fixed in #164

PR #88, opened 2022-06-16
Add thread for STM

### LINKED ISSUE #81: Add `Thread` mode for STM
For symmetry it would be natural to add a `Thread` mode for STM too, analogous to `Lin`'s `Thread` mode.

n-osborne:
should close #81 when finished


jmid:
Thanks! I agree that we are going to change the API anyway to address #66.
To minimize API changes it might indeed be better to add an STM Thread interpretation as part of that...
I'll try to elaborate a bit more in #66...

n-osborne:
I'm happy to either add this refactoring onto this PR or do it in another.

jmid:
Great - thanks :pray: - either is fine I think :+1:

lib/dune
```diff
@@ -1,8 +1,8 @@
 (library
  (name STM)
  (public_name multicorecheck.stm)
- (modules STM)
- (libraries qcheck util))
+ (modules STM StmSpec STM_Seq STM_Common STM_Domain STM_Thread)
+ (libraries threads qcheck util))
```
jmid, line 5: Here we need to express that `threads` is only a dependency of `STM_Thread` to avoid the implicit side-effects that the dependency causes (as mentioned in #66 for Lin).

Perhaps this can be done by simply splitting the `dune` rule?

Nit: module `STM` is gone, no?

jmid:
Thanks for this! :pray:  (note: we should do the same for `Lin` - but probably in a separate PR)

Perhaps it can be confusing to a newcomer that `agree_test` changes meaning across modules? :thinking: 
One possibility is to
- keep `agree_test` in the sequential functor output
- keep `agree_test_par` in the `Domain` functor output
- use `agree_test_conc` in the `Thread` functor output?

@naomi-spargo @OlivierNicole - what do you think?

We are a bit inconsistent naming-wise (`StmSpec` vs `STM_Common`).
Since the sub-modules already live inside the `STM` library an option is to omit the prefix and fully qualify the name:
`STM.Domain.Make(...)`. This comes at the risk of shadowing `Domain` and `Thread` from the `Stdlib` though.

I would love if something similar is possible for the `StmSpec.S` interface...
Renaming `StmSpec.ml` to `STM.ml` and `S` to `Spec` would yield the pleasing `STM.Spec`
with the less pleasing `STM.STM.Spec` fully qualified name :grimacing: 
I am open to suggestions to fix this... :smiley:

lib/dune
n-osborne: I've splited `STM` in several libraries so that `thread` is only a dependency of `STM_Thread`

The module `STM` is the point of entry and reexport `Spec` and the different functor.

n-osborne:
> Perhaps it can be confusing to a newcomer that `agree_test` changes meaning across modules? thinking One possibility is to
> 
>     * keep `agree_test` in the sequential functor output
> 
>     * keep `agree_test_par` in the `Domain` functor output
> 
>     * use `agree_test_conc` in the `Thread` functor output?

I was going for making the distinction in the modules' name (less function's name to find...), but I'm totally ok with your proposition.

> We are a bit inconsistent naming-wise (`StmSpec` vs `STM_Common`). Since the sub-modules already live inside the `STM` library an option is to omit the prefix and fully qualify the name: `STM.Domain.Make(...)`. This comes at the risk of shadowing `Domain` and `Thread` from the `Stdlib` though.

I agree. Also, the user shouldn't make use of `Stdlib.Domain` or `Stdlib.Thread` (which are still available fully qualified) when writing tests with STM, no? In this case, I guess shadowing is not really a big issue. Or is there a use case that I don't see?

> I would love if something similar is possible for the `StmSpec.S` interface... Renaming `StmSpec.ml` to `STM.ml` and `S` to `Spec` would yield the pleasing `STM.Spec` with the less pleasing `STM.STM.Spec` fully qualified name grimacing I am open to suggestions to fix this... smiley

The `Spec` interface is now included in the `STM` module, not in an elegant way but we have `STM.Spec` as the fully qualified name for the user. Note that the user can also access it via `STM_Core.Spec`. Not sure it is an issue?

lib/STM_Thread.ml
```diff
@@ -0,0 +1,39 @@
...
+    let obs1 = match !obs1 with Ok v -> v | Error exn -> raise exn in
+    let obs2 = match !obs2 with Ok v -> v | Error exn -> raise exn in
+    let ()   = Spec.cleanup sut in
+    check_obs pref_obs obs1 obs2 Spec.init_state
+      || Test.fail_reportf "  Results incompatible with linearized model\n\n%s"
+         @@ print_triple_vertical ~fig_indent:5 ~res_width:35
+           (fun (c,r) -> Printf.sprintf "%s : %s" (Spec.show_cmd c) (show_res r))
+           (pref_obs,obs1,obs2)
+
+  let agree_test_par ~count ~name =
```
jmid, line 31:
```ocaml
  let agree_prop_thread ...
  let agree_test_par ...
```
is a bit inconsistent with the `Domain` functor's naming.

Also, I've tried to keep the 'property' and 'test' identifiers aligned, so that they only differ in four characters (`prop` vs `test`) - but I've not written this design decision down anywhere up until now.

n-osborne:
So I've added an interface file for `STM` and cleaned up a bit the dune file.
`STM_Core` should not be shared with the world as a path (the content is still included in `STM`).

jmid:
Excellent - looks good! :+1: 

> I was also thinking of another convenience:
> One could include the output of the sequential functor in both the Domain and Thread functors.
> In that case we will need different identifiers to distinguish the different tests and properties.
> (I'd still like to keep the sequential functor though)

Any thoughts on the above convenience?
For a user only interested in, e.g., parallel `Domain` tests it would save one functor application :man_shrugging: 

As to the functor/module names, this uses a mixture of underscores `_` and CamelCase: `STM_Core`, `STM_Domain`, ...
I was just looking at https://discuss.ocaml.org/t/capitalized-underscore-vs-camelcase/502 to understand the OCaml convention.
It seems underscores are agreed upon. Since `Domain` and `Thread` refer to Stdlib module names, I can see the point in keeping them like that - and changing only one occurrence `STM_core` will then stand out and seem inconsistent, so overall I'm OK with the module naming you propose :pray: 

As to tests, it would be nice to include just one `STM_Thread` application, e.g., in `src/neg_test` to make sure that this functor indeed works as intended. Ideally the `Spec` of such a test can then live in a third file and be shared across two executables with `STM_Thread` and `STM_Domain`.

n-osborne:
Thanks :smiley:  

> > I was also thinking of another convenience:
> > One could include the output of the sequential functor in both the Domain and Thread functors.
> > In that case we will need different identifiers to distinguish the different tests and properties.
> > (I'd still like to keep the sequential functor though)
> 
> Any thoughts on the above convenience? For a user only interested in, e.g., parallel `Domain` tests it would save one functor application man_shrugging

That's true, but does a user interested in e.g. `STM_Domain` is necessarily interested in `STM_Seq`?
Also, that would meant the with both `STM_Domain.Make` and `STM_Thread.Make` built, the user would have `STM_Seq` twice.
That's not a problem because of shadowing, but I don't find it very elegant.

As a user, I don't mind applying one more functor. And w.r.t design, it's not a strong opinion, but I like the clean "one functor per kind of test" :smiley:



> As to tests, it would be nice to include just one `STM_Thread` application, e.g., in `src/neg_test` to make sure that this functor indeed works as intended. Ideally the `Spec` of such a test can then live in a third file and be shared across two executables with `STM_Thread` and `STM_Domain`.

I was waiting for the refactoring to settle a bit down before adding more Thread tests. I agree that a negative test is the bare minimum.

jmid:
I think the easiest for you would be to run `git rebase main` on your PR branch with an up-to-date `main` branch
and resolve any conflicts underway to avoid "fixing PRs".

OlivierNicole:
> @naomi-spargo @OlivierNicole - what do you think?

Sorry I hadn't reacted to this, since I get notifications from all multicoretests activity I don't always go read the messages…

lib/STM_Domain.ml
```diff
@@ -0,0 +1,36 @@
...
+
+  open Util
+  open QCheck
+  open STM_Core.Make(Spec)
+
+  (* Parallel agreement property based on [Domain] *)
+  let agree_prop_par (seq_pref,cmds1,cmds2) =
+    assume (all_interleavings_ok seq_pref cmds1 cmds2 Spec.init_state);
+    let sut = Spec.init_sut () in
+    let pref_obs = interp_sut_res sut seq_pref in
```
jmid, line 13: This seems to revert #45?

jmid: Oh, sorry - #45 was only done for `Lin`!
I suggest we do something similar for `STM` - but that it probably better done in a separate PR

jmid:
> I think the decision has been taken now?

We are still open to input to arrive at a nice interface. Your perspective is therefore welcome!
Furthermore, I expect that we will make an analogous split in Lin following this PR.

lib/STM_Core.ml
```diff
@@ -0,0 +1,281 @@
...
+
+  (* ****************************** additions from here ****************************** *)
+
+  let check_and_next (c,res) s =
+    let b  = Spec.postcond c s res in
+    let s' = Spec.next_state c s in
+    b,s'
+
+  (* operate over arrays to avoid needless allocation underway *)
+  let interp_sut_res sut cs =
```
jmid, line 99: A "default interpreter" feels a bit overkill when there are only two "backend interpreters" :thinking: 
The alternative to including a "default" interpreter targeting `Domain` and overriding it in `STM_Thread` would be to let each of `STM_Domain` and `STM_Thread` provide their own. 

Can you share your thoughts on this choice? (I might be missing something)

n-osborne: I agree this is not the best design. I was focusing on having `agree_test_conc` finding the expected bugs (which still fail).
 
There is also the re-exportation of four functions in `STM_Seq` (the reason is that they are needed in `src/lockfree/ws_deque_test.ml`) which is not a great design and is related to this one.

n-osborne:
rebased on latest main with  QCheck 0.19

jmid:
The latest run starts by triggering an error in `dune` in the CI:
https://github.com/jmid/multicoretests/runs/7279773909?check_suite_focus=true#step:6:14

I suggest
- creating a minimal reproducable test case and filing an issue on `dune`
- figuring out why it errors here (it seems related to `STM_Core`?

src/buffer/stm_tests.ml
```diff
@@ -125,12 +124,13 @@ struct
...
 ;;
 Util.set_ci_printing ()
 ;;
 QCheck_base_runner.run_tests_main
-  (let count = 1000 in
-   [BufferSTM.agree_test         ~count ~name:"STM Buffer test sequential";
-    BufferSTM.neg_agree_test_par ~count ~name:"STM Buffer test parallel" (* this test is expected to fail *)])
+  (let count = 100 in
+   [BufferSTM_Seq.agree_test     ~count ~name:"buffer test";         (* this test is expected to succeed *)
+    BufferSTM_Dom.agree_test_par ~count ~name:"buffer test parallel" (* this test is expected to fail *)])
```
jmid, line 136: I think this is reverted from a negative test

src/hashtbl/stm_tests.ml
```diff
@@ -117,12 +117,13 @@ struct
...
+module HTest_Dom = STM.Domain.Make(HConf)
 ;;
 Util.set_ci_printing ()
 ;;
 QCheck_base_runner.run_tests_main
   (let count = 200 in
-   [HTest.agree_test         ~count ~name:"STM Hashtbl test sequential";
-    HTest.neg_agree_test_par ~count ~name:"STM Hashtbl test parallel";
+   [HTest_Seq.agree_test     ~count ~name:"Hashtbl test";
+    HTest_Dom.agree_test_par ~count ~name:"Hashtbl test";
```
jmid, line 128: Also reverted from `neg_`

jmid:
The CI log reveals that this still triggers the `dune` error from above:
```
Internal error, please report upstream including the contents of _build/log.
Description:
  ("Unexpected find result", { found = Not_found; lib.name = "STM_Core" })
Raised at Stdune__code_error.raise in file "otherlibs/stdune/code_error.ml",
  line 11, characters 30-62
Called from Fiber__scheduler.exec in file "src/fiber/scheduler.ml", line 69,
  characters 8-11
-> required by ("<unnamed>", ())
-> required by ("<unnamed>", ())
-> required by ("<unnamed>", ())
-> required by ("load-dir", In_build_dir "default/doc/example")
-> required by ("toplevel", ())

I must not crash.  Uncertainty is the mind-killer. Exceptions are the
little-death that brings total obliteration.  I will fully express my cases. 
Execution will pass over me and through me.  And when it has gone past, I
will unwind the stack along its path.  Where the cases are handled there will
be nothing.  Only I will remain.
```

Also the logs reveal test failures in `Hashtbl`, `Ws_deque`, `Buffer`, ... which surprises me.
A quick look at the code diff indicates that some of these were reverted from negative `neg_` tests.

src/buffer/stm_tests.ml
n-osborne: Yes, thanks for spotting this!

src/hashtbl/stm_tests.ml
n-osborne: Again, thanks!

n-osborne:
Filled [this issue](https://github.com/ocaml/dune/issues/6153) regarding the dune bug.

n-osborne:
The "way around" the dune problem is not satisfactory, as the examples in `doc/` don't build anymore. The problem just does not show up in the CI.

Now that I understand the problem a bit more, a possible solution is to make `STM_Core` public.

jmid:
I just gave this a spin:
```
$ opam pin . -y
This will pin the following packages: multicorecheck, multicoretests. Continue? [Y/n] y
[...]

$ ocaml
OCaml version 5.0.0~alpha1
Enter #help;; for help.

Findlib has been successfully loaded.  [...]

# #require "qcheck-core";;
/home/jmi/.opam/5.0.0~alpha1/lib/bytes: added to search path
/home/jmi/.opam/5.0.0~alpha1/lib/ocaml/unix: added to search path
/home/jmi/.opam/5.0.0~alpha1/lib/ocaml/unix/unix.cma: loaded
/home/jmi/.opam/5.0.0~alpha1/lib/qcheck-core: added to search path
/home/jmi/.opam/5.0.0~alpha1/lib/qcheck-core/qcheck_core.cma: loaded
# #require "multicorecheck.stm";;
/home/jmi/.opam/5.0.0~alpha1/lib/qcheck-core/runner: added to search path
/home/jmi/.opam/5.0.0~alpha1/lib/qcheck-core/runner/qcheck_runner.cma: loaded
/home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/__private__/util: added to search path
/home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/__private__/util/util.cma: loaded
/home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/__private__/STM_Core: added to search path
/home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/__private__/STM_Core/STM_Core.cma: loaded
/home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/__private__/STM_Domain: added to search path
/home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/__private__/STM_Domain/STM_Domain.cma: loaded
/home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/__private__/STM_Seq: added to search path
/home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/__private__/STM_Seq/STM_Seq.cma: loaded
/home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/__private__/STM_Thread: added to search path
# #require "multicorecheck.stm";;multicorecheck/__private__/STM_Thread/STM_Thread.cma: loaded
Error: Reference to undefined global `Thread'
# 
```

So to me it seems this refactoring still requires loading `Thread`, which has side-effects as described in #66 that I was hoping to avoid in pure `Domain` tests (the current `Lin` structure also has this problem).

Unrelated: I see one CI failure due to a `int64 ref` failing although it was marked as positive

jmid:
I'm still a bit unsure of this and tried to dig some more...
```
ocamlobjinfo ~/.opam/5.0.0~alpha1/lib/multicorecheck/stm/STM.cmxa 
File /home/jmi/.opam/5.0.0~alpha1/lib/multicorecheck/stm/STM.cmxa
Extra C object files:
Extra C options:
Name: STM
CRC of implementation: 913d03dbd91e970e990105371b8c08f0
Globals defined:
	STM
Interfaces imported:
	20e6ebc6e7a6951b21e4e8479eeefab6	Util
	9ae42b66886b9664afb11002277f5833	Unix
	b066583d4cba6d68bde86a78d92d40ea	Thread
	9ea864085095970c73b7b17c26c37317	Stdlib__Uchar
	cc3f899885b746f777815dc93be7ede2	Stdlib__Seq
	3f2ddc04bb28ff1050c0c8d7cf5a8d57	Stdlib__Random
	8debea255efc834da5c1bc2e73f09214	Stdlib__Printf
	60fe9c49a0c9858647ea3bd6b4b90418	Stdlib__Printexc
	26fe71351d86ebe5d69a94600294de33	Stdlib__Option
	627e6e1768175d0dea895dfbcbc11457	Stdlib__Nativeint
	feb7c568830f9f239501efb83f4725e4	Stdlib__List
	7ab3f2ace09cfcc67be8a7133d18f0e2	Stdlib__Int64
	1db604c2106f996ea47a572773a5ff7f	Stdlib__Int32
	dce68268a13b78c2ac6ad553060bde69	Stdlib__Hashtbl
	add2da620cdaf3de3b32af4be4f87104	Stdlib__Gc
	50fcf20b181026bea5d7677d300e6be7	Stdlib__Format
	9ffa2cdbd0ed99dbe0e77e808eb8353a	Stdlib__Float
	b3ad0f8c65f28252059edebc03f68808	Stdlib__Either
	c61087f86bccfe329d4b904a2e7167be	Stdlib__Domain
	d5f041ccbe9783af90ca38337002fb4e	Stdlib__Complex
	a12febb22fa329dc4107399a5d5a0585	Stdlib__Bytes
	1a6020970574ac447704826c65b2ed5e	Stdlib__Buffer
	55a812d9a3c89e7a8ebb917e63e4beea	Stdlib__Bigarray
	4e17ab3de2bef18e6b5e970ca1da0ed6	Stdlib__Atomic
	46505e59a6020d9f997b8bc23d26dff6	Stdlib__Array
	66e0858dc619621d31f7ec460cfb5273	Stdlib
	be13fac5cafbe622d65c0362872f6168	STM_Thread
	fc51e209e7bd88f621ee08d8002a2d0a	STM_Seq
	33909d8dad323ad6ec33d12593708665	STM_Domain
	b40c5944ac5c0887954773fee8f07bb4	STM_Core__Spec
	399d4a5efd2212e6de079970140190b7	STM_Core__
	ad50982f1b146a63a80a0bba31a6cfbd	STM_Core
	5b9ae4be2d94512b613b8617ea95163e	STM
	fc4616701d0987e0150dc11c93292371	QCheck2
	e357977903223d2ab26c4b310368264e	QCheck
	cecfafd5c021473f5e358c96ac7502a0	CamlinternalFormatBasics
Implementations imported:
	37a153e8ed2bae828abaf8f976e4d77a	STM_Thread
	d5e7ef1a49d53089e519f8b8fedd4c06	STM_Seq
	27b6b2ee1955bb4180ee53e91db4d56b	STM_Domain
	e5dc2565ce1651773a825b3f15c4a797	STM_Core
Clambda approximation:
  _
Currying functions:
Apply functions:
Send functions:
Force link: no
```

This output seems similar to that for `lin.cmxa` in that they both import the `Thread` interface :thinking:

n-osborne:
I guess this is because `STM` still depends on `Thread` transitively (by `STM_Thread`).

Does that mean that if we want to avoid loading `Thread` when only testing for `Domains` we should make the split at the package level?

jmid:
We may indeed have to do so! :shrug: 

BTW, here's a pointer to the side-effecting code executed when loading `Thread`: https://github.com/ocaml/ocaml/blob/trunk/otherlibs/systhreads/thread.ml#L87-L98

n-osborne:
Rebased on latest main

n-osborne:
So now, it seems an stm-test that only check for Domains does not load the Thread module:

TLDR:
```
$ ocamlobjinfo _build/default/src/array/.stm_tests.eobjs/native/dune__exe__Stm_tests.cmx | grep Thread
$
```

The long version:
```
$ ocamlobjinfo _build/default/src/array/.stm_tests.eobjs/native/dune__exe__Stm_tests.cmx
File _build/default/src/array/.stm_tests.eobjs/native/dune__exe__Stm_tests.cmx
Name: Dune__exe__Stm_tests
CRC of implementation: cf2cdebfad13e7dbbc764649e4a41a70
Globals defined:
	Dune__exe__Stm_tests
Interfaces imported:
	20e6ebc6e7a6951b21e4e8479eeefab6	Util
	9ea864085095970c73b7b17c26c37317	Stdlib__Uchar
	736623667c9f391a9f8263c54ca2b70d	Stdlib__Seq
	3f2ddc04bb28ff1050c0c8d7cf5a8d57	Stdlib__Random
	8debea255efc834da5c1bc2e73f09214	Stdlib__Printf
	60fe9c49a0c9858647ea3bd6b4b90418	Stdlib__Printexc
	26fe71351d86ebe5d69a94600294de33	Stdlib__Option
	627e6e1768175d0dea895dfbcbc11457	Stdlib__Nativeint
	feb7c568830f9f239501efb83f4725e4	Stdlib__List
	7ab3f2ace09cfcc67be8a7133d18f0e2	Stdlib__Int64
	1db604c2106f996ea47a572773a5ff7f	Stdlib__Int32
	dce68268a13b78c2ac6ad553060bde69	Stdlib__Hashtbl
	add2da620cdaf3de3b32af4be4f87104	Stdlib__Gc
	5faa92c7fd1f3deb1c28248a58fc2d76	Stdlib__Format
	06dd77753bc4a71283a1ff5e476bc83b	Stdlib__Float
	b3ad0f8c65f28252059edebc03f68808	Stdlib__Either
	99f71e1737c104abbb7b5c6e445565cc	Stdlib__Domain
	71d01c936f35204d25fbdc2a38153066	Stdlib__Char
	a12febb22fa329dc4107399a5d5a0585	Stdlib__Bytes
	1a6020970574ac447704826c65b2ed5e	Stdlib__Buffer
	4e17ab3de2bef18e6b5e970ca1da0ed6	Stdlib__Atomic
	46505e59a6020d9f997b8bc23d26dff6	Stdlib__Array
	e92ed6940c016bf361e18a56415894e5	Stdlib
	de17f71dd7917470c757f9da8fc28862	STM_sequential
	678b97d9a0a27189be60bec99f65ed37	STM_domain
	5382c99a36f634b6cf8dba20db2c9713	STM_base__STM_spec
	76a0129ac025b6825bab6380d3a7e533	STM_base__STM_internal
	a43ee9e6dbfff0a2994b23b706de37c8	STM_base__
	1364d0aa7b8b98ec0762f5b5a19d0f21	STM_base
	efa2b28efe5b44f04331d2b17e90c904	QCheck_base_runner
	fc4616701d0987e0150dc11c93292371	QCheck2
	e357977903223d2ab26c4b310368264e	QCheck
	13f35ee9b3b4d4e0e8600359ad468ff4	Ppx_deriving_runtime
	7f82d1e468c33235652bc0ffb00d2123	Dune__exe__Stm_tests
	cecfafd5c021473f5e358c96ac7502a0	CamlinternalFormatBasics
Implementations imported:
	--------------------------------	Util
	f7297532faeaaff05432fb9856d59c07	Stdlib__Seq
	28fadc66828f7923b91ed371e6f568c5	Stdlib__List
	03d2f9cf224b558dea6c50f26af27ddb	Stdlib__Format
	9dace3c111d449845b3a315ca9019b3a	Stdlib__Char
	8857e884c1d7478594833c67490663ce	Stdlib__Array
	5cfd883d007b90b7229e28d55b47bea0	Stdlib
	--------------------------------	STM_sequential
	--------------------------------	STM_domain
	--------------------------------	STM_base
	6edd5303c2d4d949ebe7894fa01f33fc	QCheck_base_runner
	7ecaaad29ad7cfc18bdc6be2c2bdf62e	QCheck
Clambda approximation:
  _
Currying functions: 3 2
Apply functions: 2
Send functions:
Force link: no
```

shym:
I’m a fussy reviewer, sorry ;-)
Reviewing the global diff (I didn’t look at individual commits), I’ve only noticed slight details:
- some comments ([1](https://github.com/jmid/multicoretests/pull/88/files#diff-4228f782cabd45fb362f8f2f4aedde82b8e8498eeb04b24140349a84c311e218R90), [2](https://github.com/jmid/multicoretests/pull/88/files#diff-ba5d4e93f8a47eda11f4216e2952d78e6fb0be4af752c66a8a11cab43f90722eR1)) that should disappear
- some inconsistencies in names: I’d favour using systematically `_seq` and `_dom` suffixes for modules (instead of `_Seq` and `_Dom` which are sometimes used)
- I see a few occurrences of `STM_Seq`, `STM_Domain` and `STM_Thread` in comments; see `git d $(git merge-base main @).. | grep '\<STM_[SDT]' -B 6 --color` to locate them
- some inconsistencies in spaces and indentation (I could fix them when the PR will be otherwise ready).
Apart those tiny points, it looks very good!

shym:
If I’m not mistaken, you duplicated `print_seq_trace` and I didn’t understand why.
Otherwise all my remarks are addressed, I think. Well done ;-)

n-osborne:
You are not mistaken, thanks for spotting that and all the rest before.
Just a rebasing mistake :-)

jmid (changes requested):
I've now taken another look at this and commented on a number of smaller things.

- Since #145 the interface documentation has been pulled out into a separate `.mli`-file. In it's current state the PR reverses #145 and the STM documentation added since. I'd be grateful if the STM.mli content could be split up into `.mli`s for the proposed modules, to avoid having to redo it :pray: 

- Since the PR has been rebased a couple of times, I'm wondering whether it represents a refactoring of the latest `STM` code. The reverted comments make me suspect it is not :thinking: 

- From a user perspective, a user has to list 3 dependencies `(libraries qcheck-stm.base qcheck-stm.sequential qcheck-stm.domain)` to write an STM test running both sequentially and using `Domain`. I understand that the first holds the functor input's module interface, and that the latter two contain one "mode" each. Still, that's definitely more than the previous `(libraries STM)` - and all of these then have to explained to potential new users. Do you see any way to simplify this requirement?

n-osborne:
>     * Since #145 the interface documentation has been pulled out into a separate `.mli`-file. In it's current state the PR reverses #145 and the STM documentation added since. I'd be grateful if the STM.mli content could be split up into `.mli`s for the proposed modules, to avoid having to redo it pray

I've now restored the lost documentation and put it in the right places. This leads to spot again a few duplicated code. I also added documentation to function newly exposed (because of the splitting in different modules).

>     * Since the PR has been rebased a couple of times, I'm wondering whether it represents a refactoring of the latest `STM` code. The reverted comments make me suspect it is not thinking

I think it is good. But I'll take another look with fresh eyes.
 
>     * From a user perspective, a user has to list 3 dependencies `(libraries qcheck-stm.base qcheck-stm.sequential qcheck-stm.domain)` to write an STM test running both sequentially and using `Domain`. I understand that the first holds the functor input's module interface, and that the latter two contain one "mode" each. Still, that's definitely more than the previous `(libraries STM)` - and all of these then have to explained to potential new users. Do you see any way to simplify this requirement?

I completely agree and that was something that was bothering me. As `qcheck-stm.base` is always necessary, it could be re-exposed by the other, so the user wouldn't have to add it in its dependencies. It is not a big library, so I guess it is ok to have it twice (in `qcheck-multicoretests.sequential` and in `qcheck-multicoretsts.domain`).

n-osborne:
The divergence that appeared in one (or more) of the rebasing is now corrected.
Modifications in `main:lib/STM.ml` didn't show in `stm-thread:lib/STM_*.ml` and a diff didn't really help, so I've traced all the code movement done by this PR:

```
type 'a ty = ..                                        --> STM_spec.ml
type _ ty +=                                           --> STM_spec.ml
type 'a ty_show =                                      --> STM_spec.ml
let unit =                                             --> STM_spec.ml
let bool =                                             --> STM_spec.ml
let char =                                             --> STM_spec.ml
let int =                                              --> STM_spec.ml
let int32 =                                            --> STM_spec.ml
let int64 =                                            --> STM_spec.ml
let float =                                            --> STM_spec.ml
let string =                                           --> STM_spec.ml
let bytes =                                            --> STM_spec.ml
let option spec =                                      --> STM_spec.ml
let exn =                                              --> STM_spec.ml
let show_result show_ok show_err =                     --> STM_spec.ml
let result spec_ok spec_err =                          --> STM_spec.ml
let list spec =                                        --> STM_spec.ml
let array spec =                                       --> STM_spec.ml
let seq spec =                                         --> STM_spec.ml
type res =                                             --> STM_spec.ml
let show_res (Res ((_,show), v)) =                     --> STM_spec.ml
module type StmSpec =                                  --> STM_spec.ml (renamed as Spec)
  type cmd                                             --> STM_spec.ml
  type state                                           --> STM_spec.ml
  type sut                                             --> STM_spec.ml
  val arb_cmd : state -> cmd arbitrary                 --> STM_spec.ml
  val show_cmd : cmd -> string                         --> STM_spec.ml
  val init_state : state                               --> STM_spec.ml
  val next_state : cmd -> state -> state               --> STM_spec.ml
  val init_sut : unit -> sut                           --> STM_spec.ml
  val cleanup : sut -> unit                            --> STM_spec.ml
  val precond : cmd -> state -> bool                   --> STM_spec.ml
  val run : cmd -> sut -> res                          --> STM_spec.ml
  val postcond : cmd -> state -> res -> bool           --> STM_spec.ml
module Make(Spec : StmSpec) =                          --> STM_internal.ml
  let rec gen_cmds arb s fuel =                        --> STM_internal.ml
  let rec cmds_ok s cs =                               --> STM_internal.ml
  let arb_cmds s =                                     --> STM_internal.ml
  let consistency_test ~count ~name =                  --> STM_internal.ml
  let rec interp_agree s sut cs =                      --> STM_internal.ml
  let rec check_disagree s sut cs =                    --> STM_internal.ml
  let print_seq_trace trace =                          --> STM_sequential.ml 
  let agree_prop =                                     --> STM_sequential.ml
  let agree_test ~count ~name =                        --> STM_sequential.ml
  let neg_agree_test ~count ~name =                    --> STM_sequential
  let check_and_next (c,res) s =                       --> STM_internal.ml
  let interp_sut_res sut cs =                          --> STM_sequential.ml & STM_domain.ml (redifined)
  let rec all_interleavings_ok pref cs1 cs2 s =        --> STM_internal.ml
  let rec check_obs pref cs1 cs2 s =                   --> STM_internal.ml
  let gen_cmds_size gen s size_gen =                   --> STM_internal.ml & STM_sequential.ml
  let shrink_cmd arb cmd state =                       --> STM_internal.ml
  let rec shrink_cmd_list arb cs state =               --> STM_internal.ml
  let shrink_triple_elems arb0 arb1 arb2 (seq,p1,p2) = --> STM_internal.ml
  let shrink_triple arb0 arb1 arb2 =                   --> STM_internal.ml reexposed in STM_sequential.ml
  let arb_triple seq_len par_len arb0 arb1 arb2 =      --> STM_internal.ml reexposed in STM_domain.ml
  let arb_cmds_par seq_len par_len =                   --> STM_internal.ml reexposed in STM_domain.ml
  let agree_prop_par (seq_pref,cmds1,cmds2) =          --> STM_domain.ml
  let agree_test_par ~count ~name =                    --> STM_domain.ml
  let neg_agree_test_par ~count ~name =                --> STM_domain.ml
module AddGC(Spec : StmSpec) : StmSpec                 --> STM_Spec.ml
  type cmd =                                           --> STM_Spec.ml
  type state =                                         --> STM_Spec.ml
  type sut   =                                         --> STM_Spec.ml
  let init_state  =                                    --> STM_Spec.ml
  let init_sut () =                                    --> STM_Spec.ml
  let cleanup sut =                                    --> STM_Spec.ml
  let show_cmd c =                                     --> STM_Spec.ml
  let gen_cmd s =                                      --> STM_Spec.ml
  let shrink_cmd s c =                                 --> STM_Spec.ml
  let arb_cmd s =                                      --> STM_Spec.ml
  let next_state c s =                                 --> STM_Spec.ml
  let precond c s =                                    --> STM_Spec.ml
  let run c s =                                        --> STM_Spec.ml
  let postcond c s r =                                 --> STM_Spec.ml
```

shym:
I took @jmid’s last comment as a cue to have a fresh look at this PR ;-)

First of all, I would have preferred to see it structured into at least the following steps (of one or more commits):
1. refactoring, by splitting STM into _domain and _sequential libraries
2. adding STM_thread
3. cosmetic changes (I think I saw a couple of them on the way).

As it is not structured thus, I went on to review the global diff rather than the commits, trusting `git` to tell what is interesting from what I could skip: `git diff --color-moved=dimmed-zebra $(git merge-base main @) @`. So I only reviewed what seemed more than code shuffling. I opened a small [PR](https://github.com/n-osborne/multicoretests/pull/1) to polish it. In particular, when `qcheck-stm.sequential` or `qcheck-stm.domain` is needed, `qcheck-stm.base` is implicitly added by dune, so no need to mention explicitly, which answers one of @jmid’s remarks. Another commit my small PR fixes a few counts that were changed on `main` and not on the PR. Those were the only mishaps left I saw.

During that review I noticed another thing I’d rather see changed: some lines are really long; I’m not completely against using more than 80 columns since we seldom have so little screen space available and enforcing it can make code less legible, but I think more than, say, 140 is never legible anyway (and that’s what I see in full screen, pretty much).

jmid:
I have taken another pass over the PR, going over the differences between `shym/polish-stm-threads` and `jmid/main`.
I've then written commits to keep the diff minimal:
- removing/deleting changed white space
- keeping code changes to a minimum
- reverting a few CI tests that were enabled by the PR and could cause the CI test runs to differ more than necessary
- adjusting typos in documentation
- adjusting a few test names

The result is available here: https://github.com/jmid/multicoretests/tree/more-polish-stm-threads

You should be able to see the (slightly smaller) difference against `main` here 
(a dummy PR that displays the diff against main, so press 'Files changed' but don't press 'Create pull request'):
https://github.com/jmid/multicoretests/compare/main...jmid:multicoretests:more-polish-stm-threads?expand=1

n-osborne:
Thanks @jmid for this new this new polish. I'm totally ok for it to be merged on this PR.

jmid:
I continued by looking at the proposed STM interfaces:

`check_obs` and `shrink_triple` with the below signatures involve triples and hence morally belong in `STM_domain` rather than `STM_sequential`:
```ocaml
 val check_obs : (Spec.cmd * res) list -> (Spec.cmd * res) list -> (Spec.cmd * res) list -> Spec.state -> bool
 val shrink_triple : (Spec.state -> Spec.cmd QCheck.arbitrary) -> (Spec.state -> Spec.cmd QCheck.arbitrary) -> (Spec.state -> Spec.cmd QCheck.arbitrary) -> (Spec.cmd list * Spec.cmd list * Spec.cmd list) QCheck.Shrink.t
```

I also found `interp_sut_res` in `STM_sequential` even though it isn't used there.
```ocaml
 val interp_sut_res : Spec.sut -> Spec.cmd list -> (Spec.cmd * res) list
```
This is a bit of a historical artifact, since the sequential mode uses `check_disagree` which interleaves running of `sut` commands with checking the received results against the model (as opposed to running to completion first and then checking). Since the binding isn't relevant ATM, I suggest removing it.


The above changes are available on the `revise-stm-thread-interfaces` branch: https://github.com/jmid/multicoretests/tree/revise-stm-thread-interfaces along with
- the required adjustments to `lockfree/ws_deque_test.ml` and
- an update to an outdated comment I spotted

jmid:
I made a few more changes yesterday which I've now pushed to https://github.com/jmid/multicoretests/tree/revise-stm-thread-interfaces
- I added a short documentation header to each module
- I tried `dune build @doc` and the spotted `module Make (Spec : STM_base.STM_spec.Spec) : sig ... end` which made me simplify the type-annotation in functor arguments
- I reverted to `include Util` so that users only have to `open STM_base`
- I updated the examples in `doc/` to the new STM structure

Overall, I'm OK to merge this and then adjust it along the way.
The CI-logs reveal that bytecode fail consistently on a few `Thread`-tests.
This is to be expected at the `Thread`-mode is still flaky. I'm surprised that the `Thread` tests run as expected native code!

Questions I'm wondering about
- How do we limit the visibility of `STM_internal` for users?
- What is the advantage of having a basically empty `STM_base` with only module aliases?
- Should we inline `STM_spec` in `STM_base` avoid splitting into too many files?

I agree with @shym about squashing the commits down to a few key commits to clean up the history.
At 40+ commits spread among 3 contributors that sounds non-trivial unless we'll go for collapsing it down to one... :grimacing:

n-osborne:
> I made a few more changes yesterday which I've now pushed to https://github.com/jmid/multicoretests/tree/revise-stm-thread-interfaces
> 
>     * I added a short documentation header to each module
> 
>     * I tried `dune build @doc` and the spotted `module Make (Spec : STM_base.STM_spec.Spec) : sig ... end` which made me simplify the type-annotation in functor arguments
> 
>     * I reverted to `include Util` so that users only have to `open STM_base`
> 
>     * I updated the examples in `doc/` to the new STM structure
 
Thanks ! I was not sure about the examples in `doc/` (whether to keep the paper version or to update them).


> Overall, I'm OK to merge this and then adjust it along the way. The CI-logs reveal that bytecode fail consistently on a few `Thread`-tests. This is to be expected at the `Thread`-mode is still flaky. I'm surprised that the `Thread` tests run as expected native code!


> Questions I'm wondering about
> 
>     * How do we limit the visibility of `STM_internal` for users?

We can't, because among these users, there are `STM_sequential`, `STM_domain` and `STM_thread`. The idea was that the name should be enough to tell the (human) users that the module shouldn't be used directly.
 
>     * What is the advantage of having a basically empty `STM_base` with only module aliases?

Dune was building this module anyway (IIRC), so we've (me and @shym) added it manually and included `STM_Spec`,

>     * Should we inline `STM_spec` in `STM_base` avoid splitting into too many files?

but now I think it is a good idea to inline it rather that include it.
 
> 
> I agree with @shym about squashing the commits down to a few key commits to clean up the history. At 40+ commits spread among 3 contributors that sounds non-trivial unless we'll go for collapsing it down to one... grimacing

My part is particularly messy (tried some refactoring that didn't work w.r.t avoiding the unnecessarily load of `Stdlib.Thread`, etc...).
I can try to clean it up a bit without breaking everything.

jmid:
OK, thanks I commit the latest changes to `n-osborne/stm-thread` and let you know :+1:

jmid:
OK, I've now pushed my remaining commits from the other branch.

Is it OK with you @shym to start rebasing and squashing commits?

shym:
As I was the one suggesting rewriting history, I can do it indeed ;-)

shym:
About the visibility of `STM_internal`, we indeed cannot hide it. But I think we could use an [alert](https://v2.ocaml.org/releases/4.14/htmlman/alerts.html) to make a strong statement this should not be used lightly.

jmid:
I've taken a stab at cleaning up the history and squashing this morning.
It was not possible to split into the 3 commits suggested by @shym unless I started breaking up existing commits by @n-osborne. As this seemed a bit overkill I therefore squashed them all into a single co-authored commit.

I'll let the CI run and then plan to merge this afternoon.

jmid:
Note: One cannot simply inline `STM_spec` into `STM_base` as `STM_internal` depends on the `Spec` module type, thus creating a cycle. One would therefore have to inline both files, perhaps renaming `STM_internal.Make` to `Make_internal`.
If we want to, we can do so in a separate PR.

jmid:
The final CI tests hit weird behaviour:
- one MacOSX runner is sig KILL'ed during the install of `ppx_lib` (we've seen this before @shym haven't we?)
- another MacOSX runner timed out after 3 hours, during what I think was a `src/lockfree/ws_deque_test.ml` test run - with no output lines printed from the test in question - and with none of the previous tests taking particularly long to run... :thinking:

jmid:
In the interest of moving forward with the release I'll merge this PR.
Overall, I think we were just unfortunate with the last run - and that nothing more is broken than usual.
For example, https://github.com/jmid/multicoretests/pull/180 which builds on top of this PR has 4 successful MacOSX CI runs...
Ideally we should create issues and pursue the observed failures further:
- bytecode `Thread` tests fail as usual, since the bytecode interpreter is even less likely to interrupt `Thread` tests at the right wrong moment (I think)
- one bytecode test hit a segfault on the `Ephemeron` test on `trunk` https://github.com/jmid/multicoretests/actions/runs/3444725852/jobs/5747632136#step:8:851
- one MacOSX got killed during ppx_lib install after only 10min https://github.com/jmid/multicoretests/actions/runs/3444725851/jobs/5747632328#step:6:527
- another MacOSX run on beta1 timed out during what I think was a src/lockfree/ws_deque_test.ml
https://github.com/jmid/multicoretests/actions/runs/3444725863/jobs/5747632278#step:8:768

PR #91, opened 2022-06-29
Port old Lin tests to combinator DSL

### LINKED ISSUE #54: Port previous Lin tests to the new signature DSL
We should port the existing Lin tests to the new signature DSL from #44.
I expect that it will help identify a few missing `GenDeconstr` bindings and thus help make the DSL more feature complete.

There's an initial example in https://github.com/jmid/multicoretests/blob/main/test/lin_api.ml which corresponds to the first half of https://github.com/jmid/multicoretests/blob/main/src/atomic/lin_tests.ml modulo a `nat` generator.

jmid:
This is a ~work-in-progress~ PR to fix #54.
- For `Atomic` we simply rename the existing example
- Porting `Kcas` required extending the interface of `Lin_api` to allow users to write custom types in signatures
- There's also a port of `Lazy` with a workaround for the lack of function generators

~Missing ports:~
- ~src/neg_tests/*~
- ~src/queue~
- ~src/stack~

*(edit: missing ports in #94)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ the last call should return `2`:
  let res2 = Hashtbl.length t;;
 ```
 
-See [test/lin_api.ml](test/lin_api.ml) for another example of testing
+See [src/atomic/lin_tests_dsl.ml](src/atomic/lin_tests_dsl.ml) for another example of testing
 the `Atomic` module.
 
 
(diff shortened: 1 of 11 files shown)


jmid:
Rebased on latest main

PR #94, opened 2022-07-04
porting to dsl

n-osborne:
This PR ports some lin tests to the DSL as mentioned in #54 
This is complementary to #91 

- porting stack/lin_test to dsl
- porting queue lin tests to DSL
- porting neg_tests to DSL


jmid:
Thanks!

Nit: In #90 I tried to stream line the file names. Following its convention, the dsl-test files should be called `lin_tests_dsl.ml` :man_shrugging:

jmid:
The CI complains about missing rules: https://github.com/jmid/multicoretests/runs/7182714053?check_suite_focus=true (at the top). Looks like a few more places needs updating with the new file names.

In #91 I've gone for keeping the old `lin_tests.ml` for a start. I guess it is just to be extra careful, so that it makes it easy to switch back. As a bonus, that makes it easier to compare the two approaches to each other :man_shrugging:

jmid:
I can see the CI is still failing. This is caused by some of the negative tests not failing as expected.

I realized that I misunderstood you yesterday: A `count` of `10` may be sufficient to trigger an negative test in a couple of local runs, but other machines (and operating systems, e.g., MacOS) may require a higher `count` to consistently trigger the negative test. The consequence is visible in the CI: `All checks have failed`.

For a negative test, I thus prefer to keep a high `count`, so that we consistently fail where we are expected to fail and avoid too many "false alarms" in the CI tests.
This should not be costly, as a QCheck test stops generating additional test input when the first counterexample is found. Locally on Linux `~count:1000` will thus be no more expensive than `~count:10` when the test fails on one of the first 10 tested inputs...

src/neg_tests/lin_tests_common_dsl.ml
```diff
@@ -61,8 +61,8 @@ struct
   let init () = CList.list_init 0
   let cleanup _ = ()
   let api =
-    [ val_ "CList.add_node" CList.add_node (t @-> int @-> returning bool);
-      val_ "CList.member"   CList.member  (t @-> int @-> returning bool);
+    [ val_ "CList.add_node" CList.add_node (t @-> nat_small @-> returning bool);
+      val_ "CList.member"   CList.member  (t @-> nat_small @-> returning bool);
```
jmid, line 65: I think it is nice to keep type names (`int`, `int64`, ...) in the signatures to keep them as close to a genuine signature as possible. This is also the reason I defined `int` to something less uniform here:
  https://github.com/jmid/multicoretests/blob/main/src/hashtbl/lin_tests_dsl.ml#L12
Then we can later return and adjust/play with distributions by making changes in only one place.

src/neg_tests/lin_tests_common_dsl.ml
```diff
@@ -71,8 +71,8 @@ struct
   let init () = CList.list_init Int64.zero
   let cleanup _ = ()
   let api =
-    [ val_ "CList.add_node" CList.add_node (t @-> int64 @-> returning bool);
-      val_ "CList.member"   CList.member  (t @-> int64 @-> returning bool);
+    [ val_ "CList.add_node" CList.add_node (t @-> nat64_small @-> returning bool);
+      val_ "CList.member"   CList.member  (t @-> nat64_small @-> returning bool);
```
jmid, line 75: (same here)

src/queue/dune
```diff
@@ -4,26 +4,31 @@
...
+ (flags (:standard -w -27))
+ (libraries qcheck lin))
+
 (executable
  (name lin_tests)
  (modules lin_tests)
  (flags (:standard -w -27))
- (libraries qcheck lin)
- (preprocess (pps ppx_deriving_qcheck ppx_deriving.show ppx_deriving.eq)))
+ (libraries qcheck lin))
```
jmid, line 25: I think the preprocessing has been mistakingly removed from the old non-DSL rule.

src/stack/dune
```diff
@@ -4,26 +4,31 @@
...
+ (flags (:standard -w -27))
+ (libraries qcheck lin))
+
 (executable
  (name lin_tests)
  (modules lin_tests)
  (flags (:standard -w -27))
- (libraries qcheck lin)
- (preprocess (pps ppx_deriving_qcheck ppx_deriving.show ppx_deriving.eq)))
+ (libraries qcheck lin))
```
jmid, line 25: (same)

src/queue/dune
```diff
@@ -4,26 +4,31 @@
...
 (rule
  (alias runtest)
  (package multicoretests)
- (deps lin_tests.exe)
+ (deps lin_tests_dsl.exe)
  (action
   (progn
-   (bash "(./lin_tests.exe --no-colors --verbose || echo 'test run triggered an error') | tee lin-output.txt")
+   (bash "(./lin_tests_dsl.exe --no-colors --verbose || echo 'test run triggered an error') | tee lin-output.txt")
    (run %{bin:check_error_count} "queue/lin_tests" 1 lin-output.txt))))
```
jmid, line 34: To get a nice error message in case of failure, the string should also be updated here:
```
   (run %{bin:check_error_count} "queue/lin_tests_dsl" 1 lin-output.txt))))
```

src/neg_tests/thread_lin_tests_dsl.ml
```diff
@@ -0,0 +1,13 @@
+open Lin_tests_common_dsl
+
+(** This is a driver of the negative tests over the Thread module *)
+
+;;
+Util.set_ci_printing ()
+;;
+QCheck_runner.run_tests_main
+  (let count = 10000 in
```
jmid, line 9: I think we should be extra careful when increasing the `count`.
In this case - even with a `count` of 10000, none of these tests still trigger a failure.
Looking at the CI logs in https://github.com/jmid/multicoretests/runs/7237062976?check_suite_focus=true
this takes `(516.8 +. 534.6 +. 390.7 +. 462.0)/. 60. = 31.735` minutes - that's half an hour extra!

n-osborne: You're totally right, I was just trying to see whether a failure was finally triggered or if the problem is deeper.
The problem is definitely deeper...

n-osborne:
rebased on latest main with  QCheck 0.19

src/neg_tests/lin_tests_common_dsl.ml
```diff
@@ -0,0 +1,86 @@
...
+end
+
+module Ref_int_spec : Lin_api.ApiSpec = struct
+  type t = int ref
+  let init () = Sut_int.init ()
+  let cleanup _ = ()
+  let api =
+    [ val_ "Sut_int.get"  Sut_int.get  (t @-> returning int);
+      val_ "Sut_int.set"  Sut_int.set  (t @-> int @-> returning unit);
+      val_ "Sut_int.add"  Sut_int.add  (t @-> int @-> returning unit);
```
jmid, line 34: Haven't these two also been switched to using a uniform `int` generator?

src/neg_tests/lin_tests_common_dsl.ml
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+module Ref_int64_spec : Lin_api.ApiSpec = struct
+  type t = int64 ref
+  let init () = Sut_int64.init ()
+  let cleanup _ = ()
+  let api =
+    [ val_ "Sut_int64.get"  Sut_int64.get  (t @-> returning int64);
+      val_ "Sut_int64.set"  Sut_int64.set  (t @-> int64 @-> returning unit);
+      val_ "Sut_int64.add"  Sut_int64.add  (t @-> int64 @-> returning unit);
```
jmid, line 47: Similarly, this now uses a uniform `int64` generator, doesn't it?

src/neg_tests/lin_tests_common_dsl.ml
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+module RT_int = Lin_api.Make(Ref_int_spec)
+module RT_int64 = Lin_api.Make(Ref_int64_spec)
+
+(** ********************************************************************** *)
+(**                  Tests of the buggy concurrent list CList              *)
+(** ********************************************************************** *)
+
+let int,int64 = nat_small,nat64_small
```
jmid, line 60: One could move this line up to be in scope for all uses of `int` and `int64` in the file.
Alternatively, the definitions can be pushed into the individual `Spec` modules, so that it is clear to the reader that `int`/`int64` is overridden... :thinking:

n-osborne: I've put the definitions in the respective `Spec` module. Closer to the reason is clearer I think.

jmid:
Thanks - now merged!

PR #95, opened 2022-07-06
Reorganize and rename Lin_api combinators

jmid:
This little PR reorganizes and renames some of the `Lin_api` combinators:
- we had both `nat_small` and `small_nat` too many lines apart to notice, so remove the latter and order them thematically (`int*` together and so forth, largely moving from smaller to larger base types)
- rename `pos_int` to `int_pos`
- finally update the `Ephemeron` test to reflect the renaming


jmid:
Rebased on latest main

PR #100, opened 2022-07-14
Add Lin_api test of Bytes

jmid:
This PR contains Lin_api tests of the `Bytes` module from @naomiiiiiiiii

These are the tests that originally triggered an infinite loop on ARM64.

(The commits should probably be squashed before merging)


n-osborne:
Cool!
Looks good to me.

jmid:
Rebased on latest main and squashed

PR #101, opened 2022-08-04
Remove outdated, duplicate code in src/lockfree/ws_deque_test.ml

OlivierNicole:
We seem to have forgotten to remove some old things in `src/lockfree/ws_deque_test.ml`. It contains duplicates of functions like `agree_test`, `agree_test_par`… this PR removes them.

This PR also removes the call to `agree_test_par_negative`, whose goal is unclear to me since it is defined as the same thing as `agree_test_par`… :thinking:

### DIFF
--- a/src/lockfree/ws_deque_test.ml
+++ b/src/lockfree/ws_deque_test.ml
@@ -58,70 +58,10 @@ struct
     | _,_ -> false
 end
 
-module WSDT = STM.Make(WSDConf)
+module WSDT = STM.Make(WSDConf) ;;
 
-
-(*
-;;
-QCheck_runner.run_tests ~verbose:true [
-    WSDT.agree_test     ~count:1_000 ~name:"sequential ws_deque test";
-    WSDT.agree_test_par ~count:1_000 ~name:"parallel ws_deque test (w/repeat)";
-    agree_test_par      ~count:1_000 ~name:"parallel ws_deque test (w/non_det module)";
-  ]
- *)
-
-let agree_prop_par =
-  (fun (seq_pref,owner,stealer) ->
-    assume (WSDT.cmds_ok WSDConf.init_state (seq_pref@owner));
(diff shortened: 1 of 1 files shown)


jmid:
The definitions differ slightly, so these are not left over by accident. This has to do with how work-stealing deques are supposed to be used according to spec:
- `agree_prop_par` differs in that it only `spawn`s one domain ("a stealer domain") in parallel with the original "owner domain" (it also uses `Binary.Semaphore`)
- `arb_triple` differs in that there is a difference in what each triple component generates ("owner domain" `cmd`s can't be `steal`, "stealer domain" `cmd`s can only be `steal`)
- `agree_test_par_negative` tests usage not following the spec, hence a negative test

All this could clearly have been better explained in a comment (PR welcome).
Even better: we should add/design nice combinators to make it easier for users to generate different `cmd`s in different domains.

OlivierNicole:
OK, good to know! I'll try to do propose an explanation comment soon.



>  `agree_test_par_negative` tests usage not following the spec, hence a negative test

I understand that, but I don't really understand how `agree_test_par_negative` can be useful as it's strictly the same, definitionally, as `agree_test_par`.

jmid:
> > `agree_test_par_negative` tests usage not following the spec, hence a negative test
> 
> I understand that, but I don't really understand how `agree_test_par_negative` can be useful as it's strictly the same, definitionally, as `agree_test_par`.

It is defined like this:
```ocaml
let agree_test_par_negative ~count ~name = WSDT.agree_test_par ~count ~name
```
Hence the tested property is the standard one
- which will perform two `spawn`s
- with the `cmd` generator from `WSDConf`- which will generate `Push` and `Pop` in either domain

I admit this is all very implicit - hence would benefit from a comment.
A more general triple combinator to avoid the (almost) copy-paste would be even better.

PR #113, opened 2022-08-18
Save CI time

jmid:
The CI runs are starting to take a long time. On recent `main` I see:
- MacOS: 45m-1h 
- Linux: 1h-1h20m

This PR therefore takes a pass over all the tests triggered by `dune runtest -j1 --no-buffer --display=quiet` in an attempt to reduce the time spent.

In most cases this amounts to reducing test `count` of long-running tests.
In `Stack` and `Queue` `Lin_api` tests most time was spent on shrinking the very large `int` arguments generated - so here changing to a generator of smaller numbers should make a difference.

Fingers crossed that the CI run of this PR show some actual speed-ups... :crossed_fingers:

### DIFF
--- a/src/atomic/stm_tests.ml
+++ b/src/atomic/stm_tests.ml
@@ -71,6 +71,6 @@ module AT = STM.Make(CConf)
 Util.set_ci_printing ()
 ;;
 QCheck_runner.run_tests_main
-  (let count = 1000 in
+  (let count = 250 in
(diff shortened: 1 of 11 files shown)


jmid:
With the PR we are at:
- MacOS: 30-40min - down from 45min-1h
- Linux: 45-1h down from 1h-1h20min
which is a definite improvement.

Looking at the CI log for the longest Linux run I spotted a few more opportunities to reduce...

PR #99, opened 2022-07-13
[wip] improve bug finding with Thread

n-osborne:

PR #114, opened 2022-08-19
Reduce qcheck deps

jmid:
This PR reduces the QCheck dependencies of both opam packages from `"qcheck"` to `"qcheck-core"`.

To do so, it reduces the dependencies in 
- the opam files
- `STM`, `Lin`, and `Lin_api`
- the `src` tests
- the `doc` examples
- the frontpage README

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ end
 
 module HT = Lin_api.Make(HashtblSig)
 ;;
-QCheck_runner.run_tests_main [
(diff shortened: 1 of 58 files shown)


jmid:
Merging. The failure is due to `Ephemeron` `Thread` tests triggering a failure.
It would be great to increase the chance of it doing so, e.g., as part of #99.

Issue #56, opened 2022-04-27
Extend signature DSL with a program printer


jmid:
Each generated test input represents a program of a certain, stylistic form.
Upon failure, it would be handy to be able to print the corresponding program
- without the interpretive overhead
- without a dependency on Lin

Bonus points for printing the program to a file.

Overall, this will make it easier to recreate and share any discovered issues.

n-osborne:
I have recently given some though about this issue.

I was wondering whether we need (or would it be nice to have) a similar printer for STM? I believe this can be easily done with shared code.

I would suggest that we use [PPrint](https://github.com/fpottier/pprint), my experience with it is that it is really easy to use and build a nice representation for programs.

I'm happy to give ti a try.

PR #115, opened 2022-08-19
An initial program printer

jmid:
This PR takes a stab at adding a program printer #56.

This first version prints a corresponding program without performing a sequential consistency search.

It can be invokes as follows:
```
LIN_PRINT=1 dune exec src/hashtbl/lin_tests_dsl.exe
```


jmid:
Closing this stale PR for now.

PR #118, opened 2022-08-30
Clarify lockfree/ws_deque definitions

### LINKED ISSUE #106: Clarify how src/lockfree/ws_deque_test.ml differs from the STM definitions
The definitions differ slightly, so these are not left over by accident. This has to do with how work-stealing deques are supposed to be used according to spec:
- `agree_prop_par` differs in that it only `spawn`s one domain ("a stealer domain") in parallel with the original "owner domain" (it also uses `Binary.Semaphore`)
- `arb_triple` differs in that there is a difference in what each triple component generates ("owner domain" `cmd`s can't be `steal`, "stealer domain" `cmd`s can only be `steal`)
- `agree_test_par_negative` tests usage not following the spec, hence a negative test

All this could clearly have been better explained in a comment (PR welcome).
Even better: we should add/design nice combinators to make it easier for users to generate different `cmd`s in different domains.

_Originally posted by @jmid in https://github.com/jmid/multicoretests/issues/101#issuecomment-1206306612_

jmid:
This PR fixes #106 

Besides adding comments to clarify how the definitions differ from those in `lib/STM.ml` it also
- adds `shrink_triple` to the public interface and
- reuses it to minimize the differences

### DIFF
--- a/lib/STM.ml
+++ b/lib/STM.ml
@@ -125,7 +125,7 @@ module Make(Spec : StmSpec)
     val interp_sut_res : Spec.sut -> Spec.cmd list -> (Spec.cmd * res) list
     val check_obs : (Spec.cmd * res) list -> (Spec.cmd * res) list -> (Spec.cmd * res) list -> Spec.state -> bool
     val gen_cmds_size : Spec.state -> int Gen.t -> Spec.cmd list Gen.t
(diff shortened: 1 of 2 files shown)


jmid:
Merging as the CI failure is the known `Ephemeron` abort.

PR #121, opened 2022-09-02
Add linearizability tests for Bigarray.Array1 of ints

shym:
Rely on Lin_api to do the heavy lifting for those tests


jmid:
The CI is relatively happy (1/4 checks failed because it triggered the unrelated `Epmeron` issue), so I'll merge.

Thanks! :smiley:

PR #122, opened 2022-09-05
Add linearizability tests for Float.Array

shym:
Rely on Lin_api to do the heavy lifting for those tests


jmid:
Looks great - and the CI is happy too! 
Thanks! :pray:

PR #124, opened 2022-09-06
Add STM tests for FloatArray

Lucccyo:
Testing function from FloatArray module - sequential and parallel tests

### DIFF
--- a/src/floatarray/dune
+++ b/src/floatarray/dune
@@ -1,11 +1,25 @@
-;; Test of the Float.Array module of the standard library
+;; Test of the floatarray library
 
 ;; this prevents the tests from running on a default build
+
 (alias
  (name default)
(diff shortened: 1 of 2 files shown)


jmid:
Code looks good. The model descriptions are nice and simple as they should be :+1: 

A few things:
- the code is duplicated in both `float.array` and in `floatarray` - we should add just one copy
- minor nit: I spotted `SMT` in the comments
- finally there are conflicts which should be resolved - probably by rebasing on the latest `main` (I can help with this if you need it)

jmid:
I'm puzzled why this hasn't kicked off any CI tests :thinking:

jmid:
I have just check manually that this works fine locally on my machine, so I'll merge. Thanks :+1:

PR #126, opened 2022-09-09
Add tests for domains and threads interactions

shym:
These tests generate random trees of spawning nodes (either a domain or a thread) and then joining them in another random pattern

These tests only come with generators, not shrinkers, at the moment. It would be interesting to add some shrinkers if we run into failing examples.


jmid:
Looks good to me, thanks!

PR #128, opened 2022-09-10
Document the Util module

jmid:
Here's a PR providing documentation for the shared `Util` module.

### DIFF
--- a/lib/util.ml
+++ b/lib/util.ml
@@ -1,5 +1,3 @@
-(* Repeat a non-determistic property                          *)
-(* This is handy if the outcome depends on, e.g., scheduling. *)
 let rec repeat n prop = fun input ->
   if n<0 then failwith "repeat: negative repetition count";
   if n=0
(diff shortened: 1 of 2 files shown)


lib/util.ml
```diff
@@ -1,5 +1,3 @@
-(* Repeat a non-determistic property                          *)
```
n-osborne, line 1: To be completely exact, this repeat a property, and it is useful when the property is non-deterministic.
Not sure it is that important though :-)

lib/util.mli
```diff
@@ -0,0 +1,49 @@
+(** The Util module contains a number of reusable functions
+    handy for multicore testing. *)
+
+
+val repeat : int -> ('a -> bool) -> 'a -> bool
+(** [repeat num prop] iterates a (non-determistic) property [prop]
```
n-osborne, line 6: Same here.

n-osborne (approved):

lib/util.ml
jmid: OK, point well-taken :+1:
This comment is removed by the PR, so it is in a sense already addressed :slightly_smiling_face:

lib/util.mli
jmid: Good point :+1: 
I agree that the comment could be clearer. I've tried to rephrase it in commit 8e78a8c

PR #129, opened 2022-09-12
[WIP] Add STM tests for Sys module

Lucccyo:


src/sys/dune
```diff
@@ -0,0 +1,22 @@
+;; Test of the Sys library
+
+;; this prevents the tests from running on a default build
+
+(alias
+ (name default)
+ (package multicoretests)
+ (deps stm_tests.exe lin_tests.exe lin_tests_dsl.exe))
```
jmid, line 8: You only need `(deps stm_tests.exe)` here as there aren't test files `lin_tests.ml` or `lin_tests_dsl.ml` in this sub-directory

src/sys/dune
```diff
@@ -0,0 +1,22 @@
...
+
+(executable
+ (name stm_tests)
+ (modules stm_tests)
+ (libraries STM)
+ (preprocess
+  (pps ppx_deriving.show ppx_deriving.eq)))
```
jmid, line 15: Here you don't need to derive an `equal` function as you are only `[@@deriving show ...]` in the code:
```
(preprocess (pps ppx_deriving.show)))
```
Nicolas spotted and fixed many places where we do so yesterday in #132

src/sys/dune
```diff
@@ -0,0 +1,22 @@
...
+ (libraries STM)
+ (preprocess
+  (pps ppx_deriving.show)))
+
+(rule
+ (alias runtest)
+ (package multicoretests)
+ (deps stm_tests.exe)
+ (action
+  (run ./%{deps} --no-colors --verbose)))
```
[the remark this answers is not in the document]
Lucccyo: Okay, I fix that

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,252 @@
...
+      if mem fs (path @ [new_dir_name])
+      then fs
+      else mkdir fs path new_dir_name perm
+    | Rmdir (path,delete_dir_name) ->
+      if mem fs (path @ [delete_dir_name])
+      then rmdir fs path delete_dir_name
+      else fs
+    | Readdir _path -> fs 
+
+  let init_sut () = try Sys.mkdir (static_path / "sandbox_root") 0o777 with Sys_error _ -> ()
```
Lucccyo, line 141: Isn't it paradoxical to use `Sys.mkdir` at the very beginning of the test, whereas we are testing `Sys.mkdir` ?
Maybe it is better to call the `Sys.command` like in the 'cleanup function'.

jmid: Yes - good question! That indeed requires `Sys.mkdir` to work reasonably :smile: 

`Sys.command` could work for now. On Windows `Sys.command` is going to work less well - but we can patch that later.

jmid:
I've taken a first quick pass over the code.

There's duplicate code:
- a duplicate `p` function in line 188 + 203. I suggest pulling it out as a top-level function and to reuse it instead
- there's duplicate logic in `postcond` line 216 + 235:
  ```ocaml
  (match find_opt fs complete_path with
              | None -> true
              | Some target_fs -> not (is_a_dir target_fs)))
  ```
  rather than copy it, it would make sense to reuse it
- I think you can reuse the code in `is_existing_is_a_dir` in line 221 and `is_a_dir` in line 241 + 265 if we turn it into a function accepting a `path` argument?

There may be other occurrences, so take it as an exercise to review your own code critically and see if you can spot other occurrences or possibilities to simplify, shorten, and reuse.

Since this is a `multicoretests` PR it should run the parallel tests too now that the model is OK.
This will probably fail, because the `Sys` functions were not designed to be used in parallel, so it is fine to leave it as a *negative test* (expected to fail). We hope that parallel testing will not cause a segmentation fault though :crossed_fingers: 

In the harder end, the underlying generators could be adjusted:
- `name_gen`: For inspiration here's how QCheck generates `printable_chars` and `numeral` characters: https://github.com/c-cube/qcheck/blob/063c1d74795a24eb77fa661d218c4715382df566/src/core/QCheck.ml#L356-L366 We would like to keep the ability to generate duplicate names though.
- `perm_gen`: Currently this is just a constant. Since all operations are run as the same user, generating an owner permission may be sufficient

Finally, as of last night we now have Windows CI tests. This means some of the Unix-specific `Sys.command` operations (`rm -rf`, `touch`, ...) need adjustments. Checking `Sys.os_type` may help here.

src/sys/stm_tests.ml
```diff
@@ -156,7 +159,7 @@ struct
 
   let precond _c _s = true
 
-  let p path = static_path / "sandbox_root" / (String.concat "/" path)
+  let p path = static_path / "sandbox_root" / (List.fold_left (/) "" path) in
```
jmid, line 162: You should remove the `in` here as it is a top-level function

src/sys/dune
```diff
@@ -1,4 +1,6 @@
-;; Tests of Semaphore.Counting
+;; Test of the Sys library
+
+;; this prevents the tests from running on a default build
```
jmid, line 3: This comment should be removed since it describes an old rule which is no longer there

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,242 @@
...
+  type sut   = unit
+
+  let (/) = Filename.concat
+
+  let update_map_name map_name k v = Map_names.add k v (Map_names.remove k map_name)
+
+  let arb_cmd _s  =
+    let name_gen = Gen.(oneofl ["aaa" ; "bbb" ; "ccc" ; "ddd" ; "eee"]) in
+    let path_gen = Gen.map (fun path -> path) (Gen.list_size (Gen.int_bound 5) name_gen) in (* can be empty *)
+    let perm_gen = Gen.(oneofl [0o777]) in
```
jmid, line 31: `Gen.return 0o777` should give you a faster, constant generator

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,242 @@
...
+  let rec rmdir_model fs path delete_dir_name =
+    match fs with
+    | File _      -> fs
+    | Directory d ->
+      (match path with
+      | [] ->
+        (match Map_names.find_opt delete_dir_name d.fs_map with
+        | Some (Directory target) when Map_names.is_empty target.fs_map ->
+          Directory {d with fs_map = Map_names.remove delete_dir_name d.fs_map}
+        | None | Some (File _) | Some (Directory _) -> fs)
```
jmid, line 99: Isn't this last `Some (Directory _)` case redundant?

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,242 @@
...
+      else touch_model fs path new_file_name perm
+
+  let init_sut () =
+    match Sys.os_type with
+    (* | "Unix" -> ignore (Sys.command ("rm -rf " ^ (static_path / "sandbox_root") ^ " && mkdir " ^ (static_path / "sandbox_root"))) *)
+    | "Unix" -> ignore (Sys.command ("mkdir " ^ (static_path / "sandbox_root")))
+    (* | "Win32" ->
+      ignore (Sys.command (
+      "powershell -Command \"Remove-Item -Path " ^ (static_path / "sandbox_root") ^ " -Recurse -Force -ErrorAction Ignore \"
+       & mkdir " ^ (static_path / "sandbox_root"))) *)
```
jmid, line 152: You can safely remove these commented out lines now

src/sys/stm_tests.ml
Lucccyo: Good point !

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,242 @@
...
+      (match res with
+       | Error err ->
+         (match err with
+           | Sys_error s ->
+             (s = (p complete_path) ^ ": Permission denied") ||
+             (s = (p complete_path) ^ ": File exists" && mem_model fs complete_path) ||
+             (s = (p complete_path) ^ ": No such file or directory" && not (mem_model fs path)) ||
+             (s = (p complete_path) ^ ": Not a directory" && not (path_is_a_dir fs complete_path))
+           | _ -> false)
+        | Ok () -> not (mem_model fs complete_path) && mem_model fs path && path_is_a_dir fs path)
```
jmid, line 202: Here I would suggest putting the `not (mem_model ...)` case last - as it reads more naturally

src/sys/stm_tests.ml
Lucccyo: If I remove it, where do non_empty directories go in the pattern match ? 
In none because it is either Some ...WITH this condition
                                          or None or File..     ?

src/sys/stm_tests.ml
Lucccyo: Do the CI clean everything after a run? Imagine if a test crashes (with an assert false, for example), the filesys will not be removed. I can remove it locally, but does the CI automatically reset things?

src/sys/stm_tests.ml
jmid: Sorry, I overlooked the `when` clause! (I use them very seldomly myself)

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,237 @@
...
+    | _,_ -> false
+end
+
+module Sys_seq = STM_sequential.Make(SConf)
+module Sys_dom = STM_domain.Make(SConf)
+
+;;
+QCheck_base_runner.run_tests_main [
+    Sys_seq.agree_test     ~count:1000  ~name:"STM Sys test sequential";
+    Sys_dom.agree_test_par ~count:200  ~name:"STM Sys test parallel"
```
jmid, line 236: Very minor nit: align `~name:` under each other

jmid:
It seems there is a problem with Windows giving different error messages in `Sys_error`.

For example, `Rmdir` and `Readdir` can yield `"Invalid argument"`:
```
Messages for test STM Sys test sequential:

  Results incompatible with model

   (Touch ([], "aaa", 511)) : 0
   (Rmdir ([], "aaa")) : Error (Sys_error("D:\\a\\multicoretests\\multicoretests\\_build\\default\\src\\sys\\sandbox_root\\aaa: Invalid argument"))
```

```
Messages for test STM Sys test sequential:

  Results incompatible with model

   (Touch ([], "eee", 511)) : 0
   (Readdir ["eee"]) : Error (Sys_error("D:\\a\\multicoretests\\multicoretests\\_build\\default\\src\\sys\\sandbox_root\\eee: Invalid argument"))
```

jmid:
There's also a problem with `Mkdir` it seems (when `path` is a file - not a directory?):
https://github.com/ocaml-multicore/multicoretests/actions/runs/3756940216/jobs/6383585814#step:8:9981

Lucccyo:
```
(Touch ([], "eee", 511)) : 0
(Mkdir (["eee"], "ccc", 511)) : 
     Error(Sys_error("D:\\a\\multicoretests\\multicoretests\\_build\\default\\src\\sys\\sandbox_root\\eee\\ccc: No such file or 
     directory"))
```
I think the problem here is that on Windows, a `Mkdir` will recursively create the directories if they don't exists.

jmid:
The current error case in `postcond` reads
```ocaml
(s = (p complete_path) ^ ": No such file or directory" && not (mem_model fs path)) ||
```
As I understand it, this will only be true when `"eee"` is **not** a member of the file-system model.
But here  `"eee"` **is** a member of the file-system model - but it is a freshly `touch`ed file - not a directory as expected.

Lucccyo:
> As I understand it, this will only be true when "eee" is not a member of the file-system model.
But here "eee" is a member of the file-system model - but it is a freshly touched file - not a directory as expected.

Yes, of course, you are right! I fix that mistake.

Lucccyo:
Can I conclude that on Windows, there is no separation between `Not a directory` and `No such file or directory` ? 
I just tested both windows and Linux the succession of the touch and the mkdir command, and I handled a different error.
If that is the case, I will separate the two OS in the code to be more readable.
Instead of 
```
(s = (p complete_path) ^ ": No such file or directory" &&
       (not (mem_model fs path) || not (path_is_a_dir fs complete_path) )) ||
```

jmid:
I'm largely clueless about Windows unfortunately. Overall, I would expect that the error messages in `Sys_error` might be different on Windows compared to Linux/macOS.

It is good that you are doing experiments! :+1: I suggest first getting something to work on Windows by experiments as you are doing now -  and then afterwards we can consider cleaning up the code if it is needed.

jmid:
I think the next problem is that `Touch` works differently across Linux/macOS and Windows:
https://github.com/ocaml-multicore/multicoretests/actions/runs/3757700936/jobs/6385193698#step:8:21520

I believe that `touch` of an existing file or directory just updates the last modification time, whereas the "`Touch` implementation" on Windows will fail when called with an existing file or directory...

Lucccyo:
I did some tests on my PC. The two things seem to behave the same, except touching an existing dir_name, where windows complain.

I think I miss a case in the touch postcond. If I get back n = 1 (the file can't be created), I check weather:
- the path (without the file_name) not exists
- or the path  (without the file_name) don't reach a directory

But it can also fail if the full path ( with name of file) already exists.

Lucccyo:
I went to a wrong way by testing if n = 0. This equality simply mean that the touch command went without errors, but not really if the file was really created or not https://man7.org/linux/man-pages/man1/touch.1p.html

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,275 @@
...
+  In_channel.close ic;
+  os
+
+module Sys_seq = STM_sequential.Make(SConf)
+module Sys_dom = STM_domain.Make(SConf)
+
+;;
+QCheck_base_runner.run_tests_main [
+    Sys_seq.agree_test              ~count:1000 ~name:"STM Sys test sequential";
+    if Sys.unix && uname_os () = Some "Linux"
```
shym, line 272: I didn’t find in the PR or commit message why the test behave differently under mac and linux, so I was curious what happens.

shym (commented):
I added a bunch of comments in the code but they can probably be ignored, or addressed in a further PR.

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,275 @@
...
+    | Mkdir (path, new_dir_name, perm) ->
+      Res (result unit exn, protect (Sys.mkdir ((p path) / new_dir_name)) perm)
+    | Rmdir (path, delete_dir_name) ->
+      Res (result unit exn, protect (Sys.rmdir) ((p path) / delete_dir_name))
+    | Readdir path ->
+      Res (result (array string) exn, protect (Sys.readdir) (p path))
+    | Touch (path, new_file_name, _perm) ->
+      (match Sys.os_type with
+      | "Unix" -> Res (unit, ignore (Sys.command ("touch " ^ (p path) / new_file_name ^ " 2>/dev/null")))
+      | "Win32" -> Res (unit, ignore (Sys.command ("type nul >> \"" ^ (p path / new_file_name) ^ "\" > nul 2>&1")))
```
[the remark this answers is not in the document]
jmid: Yes, I think you are right. That was my late night attempt at silencing excessive stderr messages.
Serves we well to hack too late... :sweat_smile:

src/sys/stm_tests.ml
jmid: This was purely based on empirical observations - we have not been able to trigger when running these under Linux :man_shrugging:

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,275 @@
...
+open STM
+
+module SConf =
+struct
+  type cmd =
+    | File_exists of string list
+    | Mkdir of string list * string * int
+    | Rmdir of string list * string
+    | Readdir of string list
+    | Touch of string list * string * int
```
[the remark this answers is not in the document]
jmid: The less special-casing the better!

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,275 @@
...
+    | Directory of {perm: int; fs_map: filesys Map_names.t}
+    | File of {perm: int}
+
+  type state = filesys
+
+  type sut   = unit
+
+  let (/) = Filename.concat
+
+  let update_map_name map_name k v = Map_names.add k v (Map_names.remove k map_name)
```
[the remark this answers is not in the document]
jmid: (I don't think so)

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,275 @@
...
+*)
+  let init_sut () =
+    match Sys.os_type with
+    | "Unix" -> ignore (Sys.command ("mkdir " ^ (static_path / "sandbox_root")))
+    | "Win32" -> ignore (Sys.command ("mkdir " ^ (static_path / "sandbox_root")))
+    | v -> failwith ("Sys tests not working with " ^ v)
+
+  let cleanup _ =
+    match Sys.os_type with
+    | "Unix" -> ignore (Sys.command ("rm -rf " ^ (static_path / "sandbox_root")))
```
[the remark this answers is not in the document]
jmid: :grimacing:

shym:
As access rights are actually ignored in the tests, should we get rid of them?
(To be honest, I wondered mostly because I find annoying that the rights are displayed in decimal rather than octal in the test results :smile:)

jmid:
> As access rights are actually ignored in the tests, should we get rid of them?

That could make sense - no point in generating and representing something we do not use.
Also, I don't know Windows well enough to say if it has something similar that we can map it to.

> I find annoying that the rights are displayed in decimal rather than octal in the test results

I think we could address that by defining a `type permission = int` with a hand-written `show` function :slightly_smiling_face:

src/sys/stm_tests.ml
```diff
@@ -0,0 +1,275 @@
+open QCheck
+open STM
+
+module SConf =
+struct
+  type cmd =
+    | File_exists of string list
```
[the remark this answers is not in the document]
Lucccyo: Yes I think it is a good idea, it will be clearer.

shym:
Well, `echo . > file` was a bad idea because the error when the path leading to the file doesn’t exist is generated by the shell instead of the command itself and so cannot be silenced. Arf.
Maybe using `open_out` would be better cross-platform implementation to create a file?

jmid:
Does it help to use a hack like 9cf9719? Or is that precisely the reason why things are not silenced by it (silencing errors from the commands inside the shell vs silencing errors from the shell itself)?

shym:
That’s an idea! :+1:
I’ve been pondering whether we could get bitten by issues due to closing the process’ stdout/stderr before it has a chance to use them (and especially kill them too early). Something like the difference we can observe there:
```
# let x = Unix.open_process_full "echo . > /tmp" (Unix.environment ());;
[wait for a bit...]
# Unix.close_process_full x;;
- : Unix.process_status = Unix.WEXITED 2

# Unix.close_process_full @@ Unix.open_process_full "echo . > /tmp" (Unix.environment ());;
- : Unix.process_status = Unix.WSIGNALED (-8)
```
(a process writing on a pipe when the reader has closed it will receive `SIGPIPE`, ie -8 in OCaml).
But I think that the commands we want to invoke should be silent unless and until they already failed anyhow, so that should not change much (I have no experience with `rd` so I’m not completely sure, but I’d hope that’s true).

shym:
I’m rolling back on what I was suggesting, I think `Touch` must be properly defined first:
1. either we model our `Touch` on the behaviour of unix `touch`, namely that it succeeds (silently) on existing directories and files and creates an empty file if it doesn’t exist yet; in that case, it might be better to use different implementations for unix and windows; using `touch` on unix would make sense, and it could be mimicked by a combination `Sys.file_exists` and either a command (`echo on > ...` say) or `Out_channel.open_bin`,
2. or we define it in a way that a common implementation could be used in both unix and windows. If that’s something like `echo . > ...`, the possibility of failure should be taken into account.
I’m rather in favour of 1, because I’d like `Touch` to be really `touch` :sweat_smile:

jmid:
The overall goal is just to be able to create some files, so the tests may have something to rename, delete, etc.
I think I may have been the one suggesting `touch` to achieve that easily (before we got Windows CI support).

I'm fine with either:
1. We are already special-casing - so this boils down to adapting the Windows special case to run without noise AFAICS
2. A uniform implementation is also nice (the less special cases, the better). If that means, e.g., renaming `Touch` to `Createfile` to signal something else than Unix `touch` I would be supportive of that too.

shym:
I went full-steam headed with intensive testing campaigns after choosing that second option (I named it `Mkfile` because it turned out really similar to `Mkdir`) :smile:

I spotted two strange behaviours:
- a couple of times, for instance [here](https://github.com/shym/multicoretests/actions/runs/3904631096/jobs/6670661686#step:11:310), I see that 2 identical `rmdir`s (removing the same directory) can happen in parallel and both succeed; that might be worth investigating in the stdlib to see if there’s something fishy :thinking: 
- that [log](https://github.com/shym/multicoretests/actions/runs/3904631111/jobs/6670661999#step:11:231) I don’t understand; that reminds me a bit of the bug seen on windows, with a `readdir` succeeding on non-existent directories, but we’re on linux (bytecode) :thinking:

Apart from that, I’ve spotted a couple of error messages that were not covered (as always, differing behaviours between linux, mac and windows...). I think it’d be simpler not to try to tie too strongly messages with OS (in part because we’ll eventually want cygwin, msvc, etc.). What do you think?

jmid:
> * a couple of times, for instance [here](https://github.com/shym/multicoretests/actions/runs/3904631096/jobs/6670661686#step:11:310), I see that 2 identical `rmdir`s (removing the same directory) can happen in parallel and both succeed; that might be worth investigating in the stdlib to see if there’s something fishy thinking

I agree about using it as an opportunity to dig further :+1: 
I'm hesitant to report it as a bug though, as the parallel behaviour seems innocent.

> * that [log](https://github.com/shym/multicoretests/actions/runs/3904631111/jobs/6670661999#step:11:231) I don’t understand; that reminds me a bit of the bug seen on windows, with a `readdir` succeeding on non-existent directories, but we’re on linux (bytecode) thinking

That one seems strange indeed! The directory is there, then disappears, and finally shows up again? :thinking: 

> Apart from that, I’ve spotted a couple of error messages that were not covered (as always, differing behaviours between linux, mac and windows...). I think it’d be simpler not to try to tie too strongly messages with OS (in part because we’ll eventually want cygwin, msvc, etc.). What do you think?

It's a balance, I guess :shrug: 
I fully agree that this would be simpler to omit and that it will potentially grow further as we add more targets.
On the other hand, by matching we will also catch weird unanticipated error messages.
The sharper we formulate the properties - the stronger the testsuite :shrug: 
If it is a matter of adding a couple of extra cases I probably prefer to keep it as is for now... :thinking:

shym:
> I agree about using it as an opportunity to dig further

So I did. I wrote a simple [C program](https://github.com/shym/multicoretests/blob/parallel-rmdir/rmdirs.c) that create a new directory and tries to `rmdir` it in two threads. Running that program 1000000 times in CI, more than 950000 times both `rmdir` succeed on macOS ([log](https://github.com/shym/multicoretests/actions/runs/3913294224/jobs/6689013600)). So OCaml stdlib isn’t involved in that surprising behaviour :+1:
This leaves the question on how to integrate these tests so that macOS runners can reliably avoid this or reliably reproduce it.

> > * that [log](https://github.com/shym/multicoretests/actions/runs/3904631111/jobs/6670661999#step:11:231) I don’t understand; that reminds me a bit of the bug seen on windows, with a `readdir` succeeding on non-existent directories, but we’re on linux (bytecode) :thinking:
> 
> That one seems strange indeed! The directory is there, then disappears, and finally shows up again? :thinking:

Exactly. It looks like the directory is not really read again the second time, as if it was still in a cache? :thinking: 

> It's a balance, I guess shrug I fully agree that this would be simpler to omit and that it will potentially grow further as we add more targets. On the other hand, by matching we will also catch weird unanticipated error messages. The sharper we formulate the properties - the stronger the testsuite shrug If it is a matter of adding a couple of extra cases I probably prefer to keep it as is for now... thinking

I agree on getting those tests as precise as possible. We could refine them in a further PR maybe?

shym:
I’ve added the commits I had in mind in my branch `sys-fixes-2`:
https://github.com/Lucccyo/multicoretests/compare/sys-stm-tests...shym:multicoretests:sys-fixes-2

It seems in my last runs that simultaneous parallel `rmdir` can also succeed on windows:
https://github.com/shym/multicoretests/actions/runs/3912609422/jobs/6687463288#step:9:228
so this makes the idea of tuning the generator the trigger those cases reliably often a likely solver for negative tests on both macos and windows!

I’m left wondering about two points.

1. All the tests I could see failing on macos and windows seemed to contain the pattern of parallel `rmdir`s but quite a few of them didn’t shrink down nicely to only those calls (see for instance https://github.com/shym/multicoretests/actions/runs/3912609418/jobs/6687463298#step:9:452). Could this be improved to make it obvious that this is the root cause of the failure?

2. This bytecode failure https://github.com/shym/multicoretests/actions/runs/3912609417/jobs/6687463560#step:11:382 is weird: it is a bit hard to read but it contains this conundrum
```
                                                 |
                                   (Mkdir ([], "ddd")) : Ok (())
                                                 |
               .-------------------------------------------------------------------------------.
               |                                                                               |
 (Rmdir ([], "ddd")) : Ok (())                    (Mkdir (["ddd"], "eee")) : Error (Sys_error("_sandbox/ddd/eee: No such file or directory"))
                                                                                 (File_exists ["ddd"]) : true

```
Again, as we’ve seen before in bytecode with a `readdir` succeeding on a directory that must have been removed, it looks like functions reading the state (`readdir` and `file_exists`) see an older state??? :thinking:

jmid:
> I’ve added the commits I had in mind in my branch `sys-fixes-2`: [Lucccyo/multicoretests@sys-stm-tests...shym:multicoretests:sys-fixes-2](https://github.com/Lucccyo/multicoretests/compare/sys-stm-tests...shym:multicoretests:sys-fixes-2)

Great - we should do a PR to merge these changes

> It seems in my last runs that simultaneous parallel `rmdir` can also succeed on windows: https://github.com/shym/multicoretests/actions/runs/3912609422/jobs/6687463288#step:9:228 so this makes the idea of tuning the generator the trigger those cases reliably often a likely solver for negative tests on both macos and windows!

Cool :+1: 
 
> I’m left wondering about two points.
> 
> 1. All the tests I could see failing on macos and windows seemed to contain the pattern of parallel `rmdir`s but quite a few of them didn’t shrink down nicely to only those calls (see for instance https://github.com/shym/multicoretests/actions/runs/3912609418/jobs/6687463298#step:9:452). Could this be improved to make it obvious that this is the root cause of the failure?

This is tricky - we are already
- using an `Atomic` to start the two parallel tails simultaneously (like `Lin`) and 
-  repeating the non-deterministic, tested property multiple times - `rep_count:25 * retries:15 = 375` times during shrinking as far as I can tell! 

> 2. This bytecode failure https://github.com/shym/multicoretests/actions/runs/3912609417/jobs/6687463560#step:11:382 is weird: it is a bit hard to read but it contains this conundrum
> 
> ```
>                                                  |
>                                    (Mkdir ([], "ddd")) : Ok (())
>                                                  |
>                .-------------------------------------------------------------------------------.
>                |                                                                               |
>  (Rmdir ([], "ddd")) : Ok (())                    (Mkdir (["ddd"], "eee")) : Error (Sys_error("_sandbox/ddd/eee: No such file or directory"))
>                                                                                  (File_exists ["ddd"]) : true
> ```
> 
> Again, as we’ve seen before in bytecode with a `readdir` succeeding on a directory that must have been removed, it looks like functions reading the state (`readdir` and `file_exists`) see an older state??? thinking

This indeed looks strange! It could perhaps be explained by caching (as you have suggested previously on a related observation)

jmid:
I have written a "pair generator" here: https://github.com/ocaml-multicore/multicoretests/tree/pair-gen
All tests are green (still awaiting bytecode tests to complete - but they are not expected to fail).

The basic idea is to generate a pair `path * file_or_dir_name` together rather than generate the `path` and `name` independently from one another.

`pair_gen` chooses between generating 3 things:
- (potentially) fresh path and fresh name
- existing path and (potentially) fresh name
- existing path and existing name

In the empty initial `Directory` it is hard to return a meaningful pair, so I made `gen_existing_pair` return an `option` and handle the failure case `None` by falling back to generating something fresh instead.

I did a quick statistic sanity check: on a `Directory` containing only `foobar/` out of 1000 generated pairs, around 1/3 is `([], "foobar")`

jmid:
Rebased on latest `main` - and including the 2 commits from the `pair-gen` branch... :crossed_fingers:

jmid:
This CI run was all green, so I've removed the two commits focusing tests on `Sys` and squashed the remaining commits a bit. Suggestions for more squashing are welcome, otherwise I suggest we merge this (assuming the CI turns green :crossed_fingers:)

jmid:
I've now removed the two commits focusing these tests on `Sys` alone and squashed some of the others a bit, so this should be good to go (CI is green!).

jmid:
I just spotted some left-over commented-out code which f932eda removes.

jmid:
We've been over this repeatedly - so the time has come: merging :smiley:

shym:
Even if that PR is already merged, it’s still the best place to store the fact that I could reproduce the behaviour of two simultaneous `rmdir` of the same directory under Windows/Cygwin, using the same small C program. As it was observed in CI (so Windows+MingW), I don’t think it’s necessary to try and port the small test program to MingW, it certainly must be a behaviour of the underlying Windows syscalls.

shym:
Another follow-up: I wrote a small [test](https://github.com/shym/multicoretests/blob/96f6d46e48a7f0c446963c5dedcdf1ee036cd87a/inconsistencies.ml) to track down the weird behaviour we saw where a `mkdir` fails to create a subdirectory in a given directory X but calling `file_exists` _afterwards_ nevertheless states that X exists. We had observed this behaviour only in the bytecode version on linux, but this must be by chance since the CI (and local) runs of that test confirm that it exists also on native code. It appears only a couple of times in million, though, see this [log](https://github.com/shym/multicoretests/actions/runs/3951540442/jobs/6765523218).
(This suggests that removing a directory is not an atomic operation as far as the kernel is concerned)

shym:
All the functions in `Sys` I’ve looked at are very very thin layers over the actual syscalls. `Sys.rmdir` documentation is pretty much `rmdir()` documentation.
It seems that once again we’re just observing the behaviour of the underlying syscalls / OS: I could reproduce exactly that same behaviour with a small C reproducer: https://github.com/shym/multicoretests/blob/parallel-rmdir/rmmkdir.c (`Sys.file_exists` is doing a `stat`, so I did just that).
I can easily imagine why the `rmdir` is not atomic to avoid big locks in the filesystem, but that makes it tricky to see how we could / should handle them in our tests. (When not looking specifically for them, they seem to appear fairly rarely, too rarely I think to tune our generator to run into them by chance reliably)

shym:
According to this [article](https://www.doc.ic.ac.uk/~pg/publications/Ntzik2018Concurrent.html) and the detailed [PhD thesis](https://www.doc.ic.ac.uk/~pg/publications/Ntzik2017Reasoning.pdf) behind, POSIX doesn’t mandate directory operations to be atomic:
> POSIX does not require directory operations to have an atomic “effect” in order to allow for such implementations. [...] Note that POSIX specifies an empty directory as one that contains at most “.” and “..” ([7],XBD,3.144), thus intermediate states in which an empty directory is even emptier are allowed.

jmid:
Good detective work, sir! :+1:

shym:
As the case of having an observable intermediate state during the removal of a directory cropped up again in a CI run (see that [comment](https://github.com/ocaml-multicore/multicoretests/pull/294#issuecomment-1415740804)), I’m wondering whether we should reimplement `file_exists` as a `stat` succeeding *and* such that `st_nlink` is strictly positive. :thinking:

PR #130, opened 2022-09-13
Update src/README.md with latest additions

jmid:
This PR updates the overview in `src/README.md` to include the latest tests of `Bigarray`, `Bytes`, and `Float.Array`.


jmid:
Merging as this should be non-controversial :wink:

Issue #36, opened 2022-04-12
change arb_triple to take 3 generators as parameters


jmid:
We should change `arb_triple` to take 3 generators as parameters.

This would allow us to reuse it in `ws_deque_test`.

jmid:
Fixed by #135

PR #135, opened 2022-09-22
give each parts of the program a specific cmd generator

n-osborne:
This is an attempt to solve issue #36 with no modification of the existing tests (except for `lockfree/ws_deque_test.ml` which benefits from the change)

`STM.Make` exposes a new function `arb_cmds_par_` which is parametirized by three `Spec.cmd QCheck.arbitrary` (granted, the name is far from great).

I have also updated `arb_triple` to `arb_cmds_par` in `lockfree/ws_deque_test.ml` according to the name in `STM`.

### DIFF
--- a/lib/STM.ml
+++ b/lib/STM.ml
@@ -125,7 +125,7 @@ module Make(Spec : StmSpec)
     val interp_sut_res : Spec.sut -> Spec.cmd list -> (Spec.cmd * res) list
     val check_obs : (Spec.cmd * res) list -> (Spec.cmd * res) list -> (Spec.cmd * res) list -> Spec.state -> bool
     (* val gen_cmds_size : Spec.state -> int Gen.t -> Spec.cmd list Gen.t *)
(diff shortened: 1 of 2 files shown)


jmid:
This is great - a very welcome PR! The nice simplification to the `ws_deque` test speaks for itself :tada: :+1: 

I agree that the naming isn't optimal :shrug: 

- Would it cause any more internal changes if `gen_cmds` and `gen_cmds_size` just have their interfaces extended with the generator (thus avoiding `_gen_cmds` and `_gen_cmds_size`)? I understand and appreciate you trying not to break the existing interfaces though :+1: 
- For `arb_cmds_par_` I was trying to come up with a better name... `arb_triple`? :thinking: 
- Also, I suppose there could be a follow-up issue with shrinking if one needs custom shrinking to each of the three different kinds of `cmd`s? :thinking:

jmid:
Thanks - this is great :pray: 
I switched the argument order of `arb_triple` and merged :+1:

PR #136, opened 2022-09-22
Add STM tests for Semaphore.Counting

shym:
I wonder whether using this is making proper use of the `precond`. Without this, the tests will deadlock, as can be expected, but this prevents exercising the deadlock behaviour.

### DIFF
--- a/src/semaphore/dune
+++ b/src/semaphore/dune
@@ -1,0 +1,19 @@
+;; Tests of Semaphore.Counting
+
+;; this prevents the tests from running on a default build
+(alias
+ (name default)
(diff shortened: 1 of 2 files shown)


jmid:
Wonderful - the test code looks clean and nice! :pray: 

I agree about your observation about not testing deadlock behaviour, however
- this is in itself a nice first step of testing `Semaphore.Counting`
- that we are discussing how to test presence of deadlocks is a nice indicator that we are pushing the test boundaries! :tada: :smiley:

jmid:
CI is happy, so I'll merge :+1:

PR #138, opened 2022-09-27
RT_int64 was not used

n-osborne:


jmid:
Thanks! The `AddGC` functor is indeed a bit experimental, so it makes sense to comment out the `ref` test applications if they are not used.

PR #139, opened 2022-09-27
CI: Add tests using the bytecode target

shym:
Add workflows using the bytecode-only compiler, both α1 and trunk, on Linux


jmid:
Looks great - thanks a bunch! :tada: :pray: 
Can I ask you to add two "CI-status-buttons" in the README for these two new workflows?

jmid:
OK, we have failures and timeouts - but none of these are due to the bytecode targets, so I'll merge this.

PR #140, opened 2022-09-27
Prettify CI logs

shym:
Enable colors and reduce the noise in the logs


jmid:
Great to see colors in the CI-logs - makes them more easy to scan for errors (look for red 'x'...)! :smiley: 
Also, bumping the timing to 60s between updates cleans them even more up.
I'll just let the last Linux CI finish before merging...

PR #141, opened 2022-09-28
Remove timeout in the domains+threads tests

shym:
Now that the generated join patterns cannot (should not) create deadlocks, remove the timeout so that the property won’t fail just because it ran out of time


jmid:
I'll merge as we suspect the timeout'ing CI run to be caused by a (known?) compiler/run-time defect

PR #142, opened 2022-09-28
Use the Timeout exception in fork_prop_with_timeout

shym:
Adopt in fork_prop_with_timeout the same behaviour as in prop_timeout, namely raising a Timeout exception when the child did end on a SIGALRM signal


jmid:
Thanks! This is clearly an oversight - I even documented this intended behaviour in the `.mli` :sweat_smile:

PR #143, opened 2022-09-29
STM cleanup not always run in sequential mode

jmid:
@Lucccyo and I found a situation where the `cleanup` function would not always be run in STM's sequential test mode :grimacing: 
In particular this could happen if `postcond` would raise an exception, e.g., by using `assert`.

- This PR's solution is to wrap a suitable exception handler around the sequential `agree_prop`.
- There's also an internal test case to ensure this for both the sequential and parallel modes.
- Finally I've renamed `src/internal/cleanup.ml` to `src/internal/cleanup_lin.ml` to distinguish the two `cleanup` tests


jmid:
CI failures are unrelated (Thread tests that fail to trigger issues), so I'll merge this one.

jmid:
Dang. This had merged conflicts. Now rebased on main and conflicts resolved. Taking a last round of CI before merging.

jmid:
CI is happy modulo 3 bytecode runs failing to trigger the `Lin` `int64 ref` issue as usual. Merging.

jmid:
This PR caused an additional confusing line `Agreement check raised ...` to be printed for the `User_fail` exception, which is triggered when there is simply disagreement with the model:
```
Messages for test STM Array test sequential:

  Agreement check raised exception: qcheck: user fail:
  Results incompatible with model

   (Get 6) : Ok ('a')

   (Get 6)
```

I'll therefore revert the PR. 
Instead we should recommend users to write `postcond` purely, as documented in #170

PR #145, opened 2022-09-30
Add documentation for the STM module

jmid:
This PR documents the `STM` module in a corresponding `STM.mli` file.

The majority of the documentation is simply moved from `STM.ml`.


jmid:
I'll merge as this initial STM documentation is a reasonable starting point which we can improve in later PRs...

PR #147, opened 2022-10-04
CI: Set up a 3-hour timeout on native runs

shym:
Tests on native builds should not run for more than 2 hours so a 3-hour timeout will abort failing tests sooner.
Tests on bytecode builds run often above 4 hours so keeping the default 6-hour timeout makes sense


jmid:
LGTM, thanks!

PR #148, opened 2022-10-07
CI speedup

shym:
Set up two simple ways to speed up the initialization part of CI builds: activate `dune` cache and bypass useless `brew` installations on macos.


shym:
Superseded by #157, closing.

PR #154, opened 2022-10-12
Switch CI tests to 5.0.0~beta1

jmid:
This PR switches the CI tests to the newly minted 5.0.0~beta1 release


jmid:
CI looks reasonably happy:
- the 3 byte code failures are due `Lin ref int64 test with Thread` failing to trigger. This is part of a bigger problem that `Thread` tests are currently harder for us to trigger than `Domain tests`
- the MacOSX 5.0.0~beta1 failure is because the known `Ephemeron` issue wasn't triggered on this particular run.

PR #157, opened 2022-10-14
CI improvements

shym:
Those commits should improve CI by bringing some speed-up and picking more relevant trigger events.

### DIFF
--- a/.github/bin/brew
+++ b/.github/bin/brew
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+# Dummy `brew`, to skip some useless installations
+echo Dummy: brew "$@"
+exit 0
--- a/.github/bin/sudo
+++ b/.github/bin/sudo
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+# Dummy `sudo`, to skip some useless installations
+if [ "$1" = "apt-get" -a "$2" = "install" ]; then
+  echo "Dummy sudo: installing only 'bubblewrap' only instead of running: $@"
+  /usr/bin/sudo apt-get install bubblewrap
+else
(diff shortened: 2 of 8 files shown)


jmid:
The MacOSX job being killed while installing dependencies seems weird, doesn't it? :thinking:

shym:
Yes, especially since it is sent SIGKILL. I can only imagine an issue on the runner.

jmid:
Can I ask you to retrigger the CI?

I'm probably overly careful, but since this fiddles to avoid installing MacOSX dependencies and 50% (1/2) of the MacOSX runs fail afterwards,a few more MacOSX successful CI runs would help confirm that the kill was indeed an independent failure.

shym:
I triggered the CI again on my fork. See the CI results for commit [#ca12fc5](https://github.com/shym/multicoretests/commit/ca12fc58203b4ff336583a92a066288b185846e1): native runs succeed. One really strange SEGV for bytecode `neg_tests/effect_lin_tests_dsl` in that [run](https://github.com/shym/multicoretests/actions/runs/3264074943/jobs/5364142233). But I think that’s fine as far as that PR is concerned.

jmid:
OK, thanks :pray: I've just merged a couple of PRs, so I'll let them finish and then merge this one tomorrow.

The segfault on the Effect-test in bytecode sounds fishy.
I would definitely try to reproduce as it sounds like a plausible bug :thinking:

shym:
Unfortunately I didn’t manage to reproduce the segfault locally, but on the 3rd or 4th re-run in CI, I got another SEGV on that test.
https://github.com/shym/multicoretests/actions/runs/3291783283/jobs/5426364612

PR #160, opened 2022-10-19
Reduce frequencies of operations creating Bigarrays

shym:
Tune the frequencies of Bigarray functions so that the generation of a new array is tried less often: we expect a failing test that arise from race conditions, so reducing the frequency of operations that generate new arrays (and take more resources) helps to trigger the failure

### DIFF
--- a/src/bigarray/lin_tests_dsl.ml
+++ b/src/bigarray/lin_tests_dsl.ml
@@ -42,20 +42,20 @@ struct
   let int = int_small
 
   let api =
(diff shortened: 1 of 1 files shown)


shym:
This modification was done seeing multiple failures on macos on Bigarrays. I could access a mac, reproduce the failures and check that this tuning seemed to fix the issue.

jmid:
OK, thanks - makes sense :+1:

PR #163, opened 2022-10-21
CI: Enable opam sandboxing back

shym:
Since we see a few runs failing on macOS after merging #157, try to fix the problem by enabling back sandboxing. In the failing runs, no dune cache was found so it should probably not be involved (though enabling dune cache triggers its installation early on in the CI run; could this change how other dependencies are built? :thinking:)


jmid:
Thanks! :pray: 
Both the MacOSX runs succeed with this one. I'll merge and watch the next two MacOSX runs triggered... :crossed_fingers:

jmid:
Looks good - no SIGKILL on MacOSX :+1: - thanks!

PR #164, opened 2022-10-21
Split Lin package

shym:
Refactor the Lin package into sub-libraries so that users can choose to use only the domain, the thread or the effect implementation. That way it is possible to test the domain and effect implementation in runs _without_ running first the thread initialisation code.
Put the part of the interface that should not be used by external users into `_internal` modules. They can’t be really hidden (by using dune `private_modules` stanza for instance) without breaking the test suite.

This PR is really only about the couple of commits at the tip of that branch. They are currently based off an integration branch because it is meant to be merged when #112 will have been merged.


jmid:
I've now taken a pass over this PR.
Overall I was very pleased to see how systematic it approaches the refactoring.
It made reading the diff commit-by-commit a pleasure! :smiley: 

A few things I noticed - only related to naming really:
- Why change `ApiSpec` to `Spec`? I think the former is more descriptive. `InterfaceSpec` may be more in line with established OCaml terminology though...
- I'm not sure about the `ExpandedSpec` name as it is not very descriptive: such modules contain generators, printers, shrinkers, etc., so they are further from a 'specification'. I think of this old `Lin` as an internal test representation myself. Could `Lin_internal.TestRepr` and `Lin_common.Make/Build_test_repr` work? (other name suggestions welcome!)
- Finally the old `"Lin_api"` and `"Lin"` test name prefixes should probably be updated in a separate commit of the PR

shym:
It also does not yet update the documentation, but I suggest we do this also in a separate PR. Since this one is modifying so many source files in our test suite, we would avoid a lot of conflicts by merging soon.

src/neg_tests/domain_lin_tests.ml
```diff
@@ -5,7 +5,7 @@ open Lin_tests_common
...
 QCheck_base_runner.run_tests_main
   (let count = 15000 in
-   [RT_int.neg_lin_test    `Domain ~count ~name:"Lin ref int test with Domain";
-    RT_int64.neg_lin_test  `Domain ~count ~name:"Lin ref int64 test with Domain";
-    CLT_int.neg_lin_test   `Domain ~count ~name:"Lin CList int test with Domain";
-    CLT_int64.neg_lin_test `Domain ~count ~name:"Lin CList int64 test with Domain"])
+   [RT_int_domain.neg_lin_test    ~count ~name:"Lin_api ref int test with Domain";
+    RT_int64_domain.neg_lin_test  ~count ~name:"Lin_api ref int64 test with Domain";
+    CLT_int_domain.neg_lin_test   ~count ~name:"Lin_api CList int test with Domain";
+    CLT_int64_domain.neg_lin_test ~count ~name:"Lin_api CList int64 test with Domain"])
```
jmid: These are not `Lin_api` tests so I think this test renaming is a mistake.

In the "old terminology":
- `domain_lin_tests.ml` depend on `lin_tests_common.ml` and are `Lin` tests
- `domain_lin_tests_dsl.ml` depend on `lin_tests_common_dsl.ml` and are `Lin_api` tests

I think this is caused by the confusing test file names, hence I renamed them in #197.
With the new file names
- the former would share a common `lin_tests_` prefix
- the latter would share a common `lin_tests_dsl_` prefix

src/neg_tests/domain_lin_tests.ml
```diff
@@ -5,7 +5,7 @@ open Lin_tests_common
...
 QCheck_base_runner.run_tests_main
   (let count = 15000 in
-   [RT_int_domain.neg_lin_test    ~count ~name:"Lin_api ref int test with Domain";
-    RT_int64_domain.neg_lin_test  ~count ~name:"Lin_api ref int64 test with Domain";
-    CLT_int_domain.neg_lin_test   ~count ~name:"Lin_api CList int test with Domain";
-    CLT_int64_domain.neg_lin_test ~count ~name:"Lin_api CList int64 test with Domain"])
+   [RT_int_domain.neg_lin_test    ~count ~name:"Lin DSL ref int test with Domain";
+    RT_int64_domain.neg_lin_test  ~count ~name:"Lin DSL ref int64 test with Domain";
+    CLT_int_domain.neg_lin_test   ~count ~name:"Lin DSL CList int test with Domain";
+    CLT_int64_domain.neg_lin_test ~count ~name:"Lin DSL CList int64 test with Domain"])
```
jmid, line 11: As commented on the other commit, I believe this renaming is a mistake

src/neg_tests/domain_lin_tests.ml
shym: Well spotted, sorry about that. I’m afraid it’s probably more due to an ill-advised copy-paste when cherry-picking that commit on `main`.

src/neg_tests/domain_lin_tests.ml
shym: Fixed on the way.

shym:
And rebased on `main`.

src/neg_tests/lin_tests_thread_ref.ml
```diff
@@ -2,12 +2,15 @@ open Lin_tests_common
...
 ;;
 if Sys.backend_type = Sys.Bytecode
 then
   Printf.printf "Lin ref tests with Thread disabled under bytecode\n\n%!"
 else
 QCheck_base_runner.run_tests_main
   (let count = 1000 in
-   [RT_int.lin_test       `Thread ~count ~name:"Lin ref int test with Thread"; (* unboxed, hence no allocations to trigger context switch *)
-    RT_int64.neg_lin_test `Thread ~count:15000 ~name:"Lin ref int64 test with Thread"])
+   [RT_int64_thread.lin_test     ~count ~name:"Lin ref int test with Thread"; (* unboxed, hence no allocations to trigger context switch *)
```
jmid, line 15: I think this was changed to an `int64` test by mistake - which explains why this test started failing consistently in the CI.

src/neg_tests/domain_lin_tests.ml
jmid: Fixed in 5b0d03d

src/neg_tests/lin_tests_thread_ref.ml
jmid: Fixed in 5b0d03d

jmid:
Thanks a bunch for this :pray: 
I went over the commits again, found a small nit, committed a fix, and merged into `main`

PR #169, opened 2022-10-26
CI: Share the common part of every CI workflow

shym:
These commits introduce a `common.yml` workflow with the few parameters that changes from one actual workflow to the other. It makes their general maintenance easier.
While we’re at it, now that new options are shared between all workflows, add the following options:
- possibility to pick a specific commit of the compiler
- possibility to run (various times if wanted) one specific test with one specific seed, that makes it faster to check bugs for reproducibility.
The first commit is a worthy addition all by itself, I might have added too many options afterwards (even if I found them useful).

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -15,9 +15,21 @@ on:
         description: 'Timeout'
         type: number
         default: 180
+      only_test:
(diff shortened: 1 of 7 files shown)


jmid:
I am 100% for the sharing to avoid workflow duplication.
I also agree about perhaps going a bit overboard with options - but I agree that it should be OK to include them since they have been useful to you :+1:

PR #170, opened 2022-10-28
Document expected purity in STM

jmid:
This little PR adds documentation to STM to stress which functions of the user-supplied module are expected to be pure.


jmid:
The CI failures are unrelated:
- For bytecode it was an expected `Thread` failure that didn't trigger
- For MacOSX it was an expected `Ephemeron` failure that didn't trigger

jmid:
Rebased on `main` after STM refactorings.

PR #171, opened 2022-10-28
Revert "STM cleanup not always run in sequential mode"

jmid:
Reverts jmid/multicoretests#143


jmid:
For future reference: The CI failure on MacOSX is because that particular run triggered the known `Buffer` crash.

PR #177, opened 2022-11-09
Scope-preserving shrinker for multiple-ts

jmid:
This PR implements a shrinker for `multiple-ts`.
It does so by not removing any `t` variable that later commands may depend on - hence scope preservation.
Overall we replace the built-in `list` shrinker with a custom, `env`-aware shrinker. Like for the `QCheck` `list` shrinker we use the heuristic of reducing the list spine first and individual list elements (cmd arguments) second.

The spine shrinker tries
- one element after another - hence abandoning any clever bisection which shouldn't matter greatly for lists of length at most 20 `cmd`s long
- to remove non-`t` producing `cmd`s
- to remove `t`-producing `cmd`s if they are not used in the enclosing program
- to shrink `cmd` lists starting from the back - from the rationale that later `t`s are less likely to be used as fewer `cmd`s follow and can hence depend on them. This can provide a trickling effect that allow an earlier `t` to be removed in the next round, ...

The shrinker of individual `cmd`s
- first reduces `t` variables to an earlier valid index in the `env`
- secondly reduces non-`t` arguments, thereby avoiding, e.g., having to go through several `string` shrink attempts before getting to a `t` reduction if they were interleaved

I've gone over the latest CI logs and the shrinker seems to work reasonably.

### DIFF
--- a/lib/lin.ml
+++ b/lib/lin.ml
@@ -7,7 +7,6 @@ module Var : sig
   val reset : unit -> unit
   val pp : Format.formatter -> t -> unit
   val show : t -> string
-  val shrink : t Shrink.t
 end =
 struct
   type t = int
@@ -18,7 +17,6 @@ struct
     (fun () -> counter := 0)
   let show v = Format.sprintf "t%i" v
   let pp fmt v = Format.fprintf fmt "%s" (show v)
-  let shrink = Shrink.int
 end
 
 module Env : sig
@@ -30,16 +28,8 @@ struct
   type t = Var.t list
   let gen_t_var env = Gen.oneofl env
   let valid_t_vars env v =
-    if List.mem v env then Iter.return v else Iter.return 0
-(*
     if v = 0 then Iter.empty else
-    match List.filter (fun v' -> v' <= v) env with
-    | v' :: _ when v' = v -> Iter.return v
-    | env'                ->
-      let a = Array.of_list env' in
-      let length = Array.length a in
-      Iter.map (fun i -> a.(length - i - 1)) (Shrink.int length)
-*)
+      Iter.of_list (List.rev (List.filter (fun v' -> v' < v) env))
 end
 
 module type CmdSpec = sig
@@ -57,18 +47,13 @@ module type CmdSpec = sig
       It accepts a variable generator and generates a pair [(opt,cmd)] with the option indicating
       an storage index to store the [cmd]'s result. *)
 
-  val shrink_cmd : cmd Shrink.t
+  val shrink_cmd : Env.t -> cmd Shrink.t
   (** A command shrinker.
-      To a first approximation you can use [Shrink.nil]. *)
+      It accepts the current environment as its argument.
+      To a first approximation you can use [fun _env -> Shrink.nil]. *)
 
-  val fix_cmd : Env.t -> cmd -> cmd Iter.t
-  (** A command fixer.
-      Fixes the given [cmd] so that it uses only states in the given
-      [Env.t]. If the initial [cmd] used a state that is no longer
-      available, it should iterate over all the lesser states
-      available in [Env.t]. If all the necessary states are still
-      available, it should generate a one-element iterator.
-      Can assume that [Env.t] is sorted in decreasing order. *)
+  val cmd_uses_var : Var.t -> cmd -> bool
+  (** [cmd_uses_var v cmd] should return [true] iff the command [cmd] refers to the variable [v]. *)
 
   type res
   (** The command result type *)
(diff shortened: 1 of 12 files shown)


lib/lin.ml
```diff
@@ -150,45 +122,71 @@ module MakeDomThr(Spec : CmdSpec)
...
+    map (fun seq -> (seq,p1,p2)) (shrink_cmd_list_spine seq (fun var -> cmds_use_var var p1 || cmds_use_var var p2))
     <+>
-    (match p1 with [] -> Iter.empty | c1::c1s -> Iter.return (seq@[c1],c1s,p2))
+    map (fun p1 -> (seq,p1,p2)) (shrink_cmd_list_spine p1 ctx_doesnt_use_var)
     <+>
-    (match p2 with [] -> Iter.empty | c2::c2s -> Iter.return (seq@[c2],p1,c2s))
+    map (fun p2 -> (seq,p1,p2)) (shrink_cmd_list_spine p2 ctx_doesnt_use_var)
     <+>
-    concat_map (fun p1 -> Iter.map (fun p1 -> (seq,p1,p2)) (fix_cmds seq_env p1)) (Shrink.list_spine p1)
+    (fun yield -> (match p1 with [] -> Iter.empty | c1::c1s -> Iter.return (seq@[c1],c1s,p2)) yield)
```
shym, line 166: On that line, I’m not sure I see the point of η-expansion: it’s just delaying the `match` and `@` which will happen anyway, isn’t it? I would just drop it in the name of simpler code.

lib/lin.ml
```diff
@@ -150,45 +122,71 @@ module MakeDomThr(Spec : CmdSpec)
...
     <+>
-    concat_map (fun p2 -> Iter.map (fun p2 -> (seq,p1,p2)) (fix_cmds seq_env p2)) (Shrink.list_spine p2)
+    (fun yield -> (match p2 with [] -> Iter.empty | c2::c2s -> Iter.return (seq@[c2],p1,c2s)) yield)
```
shym, line 168: Same comment here.

lib/lin_api.ml
```diff
@@ -256,10 +254,10 @@ module MakeCmd (ApiSpec : ApiSpec) : Lin.CmdSpec = struct
                   Iter.map (fun args -> Args.Fn (a,args)) (gen_shrinker_of_desc fdesc_rem args)
                        | _ -> failwith "Fn/None: should not happen")
            | Some shrk ->
-               Iter.(function (Args.Fn (a,args)) ->
-                   (map (fun a -> Args.Fn (a,args)) (shrk a))
-                   <+>
+               Iter.(function (Args.Fn (a,args)) -> (* heuristic: shrink last args first *)
```
shym, line 257: Is there a best choice for the first argument to shrink?
(My intuition would be that the best choice will change from one function to another.)

src/bytes/lin_tests_dsl.ml
```diff
@@ -8,6 +8,7 @@ module BConf = struct
 
   open Lin_api
   let int = nat_small
+  let string = string_small
```
shym, line 11: Is this related to the rest of that commit?

shym:
Very nice! `cmd_uses_var` is a much nicer interface that `fix_cmd`. :+1:
(I’ve added a few minor remarks on the way.)

lib/lin.ml
jmid: It is indeed delaying the `match` and `@` computation :+1:
OCaml evaluates its arguments eagerly - also for the infix `<+>` parameters.
Omitting the eta-expansion adds needless work to the shrinker runs that successfully reduce before getting to these lines:
```ocaml
# Iter.find (fun _ -> true) Iter.(of_list [0;1] <+> (Printf.printf "boom!\n%!"; of_list [2;3]));;
boom!
- : int option = Some 0
# Iter.find (fun _ -> true) Iter.(of_list [0;1] <+> fun yield -> (Printf.printf "boom!\n%!"; of_list [2;3] yield));;
- : int option = Some 0
```
I agree that the code would read simpler without it though. I suggest we add a comment to explain why it is there.

lib/lin_api.ml
jmid: I agree, this is very much a heuristic. Tracing shrink-logs I encountered countless lines of
```ocaml
Bytes.blit_string "loooooooooooooooooooong-bytes-strink-taking-forever-to-reduce-char-by-char" large_pos bytes_t large_pos large_length
```
where the shrinker would spend many minutes getting the first argument shrunk to `"aaaaaaaaaaaaaaa....aaaa"`, to finally get to reducing the later `int` parameters - which triggered a gradual reduction of the `string` length to just `"aaaa"` or so!

From that I inferred that in practice it is probably more common to have such "index" parameters later in an argument list (`Array.set`, `List.nth`, ...). This seems to work reasonably well in practice - and was motivated by the above - but I don't claim any optimality. Of course the heuristic is trivially defeated by swapping argument order.
Finally I would argue that left-to-right is just as much of a heuristic as right-to-left :smile:

This makes me wonder whether we can phrase an 'alternating' shrinker that, that repeatedly swaps the `Iter.t` to try candidates from... :thinking:

src/bytes/lin_tests_dsl.ml
jmid: Early on I spotted that we were generating long `string`s in the `bytes` tests - worst case, up to length 10000 (arising from `Gen.nat`). The change alters that to `string`s of up to length 100 (arising from `Gen.small_nat`), which is more in line with our other tests that generate `string`s.

lib/lin.ml
jmid: I've added a comment in 94649b9

lib/lin_api.ml
shym: An alternating shrinker would be nice indeed! At least I would be curious to see if that matters ;-)

lib/lin.ml
shym: Ah! Indeed, shrinking will drop the tail as soon as one element does succeed. On first reading, I assumed we’ll eventually get to that item anyway. Thanks for the clarification.

PR #179, opened 2022-11-10
Add CI workflows on Windows

shym:
This is a lot more complex than it should be because:
- even with a `override` opam repository providing packages for the OCaml 5 compiler with build instructions compatible with Windows, setup-ocaml fails
- many different attempts to build the compiler in a simpler way failed without many clues to help (probably because of issues with the PATH, which sh is found, or not, to call configure, etc. etc.)
- ocaml-config.3 (upstream opam package) fails to install claiming the hashes in the opam package are invalid (?), so the `override` opam repository distribute that same package without the hashes
- ocamlfind.1.9.5 as provided by the `default` opam repository doesn’t build with OCaml 5 since the Windows patch uses deprecated functions

PR #188, opened 2022-11-14
Add missing test aliases

### LINKED ISSUE #184: CI testing and compilation is interleaved
I created a branch to see if I could trigger a MacOSX CI timeout while printing the commands that dune executes underneath the hood:
https://github.com/jmid/multicoretests/tree/ci-debug-attempt

As a side-effect I learned that despite running `dune build` before `dune runtest` in the CI workflow
https://github.com/jmid/multicoretests/blob/ci-debug-attempt/.github/workflows/common.yml#L78-L82
it appears as if `ocamlopt` is still sometimes interleaved with test runs.
https://github.com/jmid/multicoretests/actions/runs/3446769124/jobs/5752070142
https://github.com/jmid/multicoretests/actions/runs/3446769126/jobs/5752070129

It may be that not all executables are correctly "aliased" (in dune terminology) :thinking:

jmid:
This PR
- adds missing test directory aliases to `src/dune`
- adds default aliases so that we build **all** tests by default (to ensure against breakage)
- annotates aliases that are currently not run, for future reference (e.g., if we want to have multiple aliases)

Overall this should fix #184 

I'm wondering if not having to interleave compiler invocations and test runs will also have an effect on #179 (CC @shym)

### DIFF
--- a/src/array/dune
+++ b/src/array/dune
@@ -6,7 +6,7 @@
  (package multicoretests)
  (deps
    stm_tests.exe
-   lin_tests.exe
+   lin_tests.exe ;; currently not run on CI
    lin_tests_dsl.exe))
 
 (executable
--- a/src/atomic/dune
+++ b/src/atomic/dune
@@ -4,7 +4,10 @@
 (alias
  (name default)
  (package multicoretests)
- (deps stm_tests.exe lin_tests.exe lin_tests_dsl.exe))
+ (deps
+   stm_tests.exe
+   lin_tests.exe ;; currently not run on CI
+   lin_tests_dsl.exe))
 
 
 ;; STM_sequential and STM_domain test of Atomic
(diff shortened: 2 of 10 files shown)


jmid:
To confirm that this indeed fixes the interleaved `dune` compilation commands in #184 I merged this repo into https://github.com/jmid/multicoretests/tree/ci-debug-attempt which has finished its two workflow runs.

In both cases there are no more `ocamlopt` commands printed and triggered underway:
- https://github.com/jmid/multicoretests/actions/runs/3461360752/jobs/5778869434
- https://github.com/jmid/multicoretests/actions/runs/3461360747/jobs/5778869414

jmid:
CI logs look OK, so I'll merge. There's
- a Linux `5.0.0+trunk` `Ephemeron` segfault
- two bytecode `Thread` failures
- a bytecode run timing out

PR #189, opened 2022-11-14
Remove STM_spec.AddGC functor

### LINKED ISSUE #178: Remove AddGc functor in STM
For the forthcoming initial release we should consider removing the `AddGc` functor from STM.
It was an early experiment to see whether having explicit GC calls interleaved would increase the chance of triggering issues. 
However it doesn't seem to be the case, nor is the `AddGc` feature used in any of our own STM tests.

We should probably keep the feature on a separate branch in case it will be needed later...

jmid:
This PR
- removes the `STM_spec.AddGC` functor and 
- cleans up the old commented out tests

We thereby fix #178 

For future reference the removed code is available on the `stm-addgc` branch: https://github.com/jmid/multicoretests/tree/stm-addgc


jmid:
CI error summary
- two `5.0.0+trunk` `Ephemeron` segfaults
- two bytecode, including 3 `Thread` failure-to-trigger-issues and 1 `Ephemeron` segfault in `5.0.0+trunk` bytecode mode
- two bytecode timeouts, including 1 `STM` `Thread` failure-to-trigger-issue

I will therefore merge this.

PR #190, opened 2022-11-15
Remove a weird unicode character

shym:
I wonder how it came to appear, but anyway, it should go away.


jmid:
I think I am to blame from my rebasing of #170 after merging #88 :grimacing:

PR #191, opened 2022-11-15
Expand on STM's postcond state parameter

### LINKED ISSUE #102: Clarify in the STM docs that postcond is meant to test the pre-command state, not the post-command one
This week @ElectreAAS attempted to use STM to test her lock-free map, and we were both as confused with the results.

It turns out that we had a wrong `postcond`: we tried to check post-conditions on the state *after* running the command, whereas `postcond` can only check things on the input state.

This is stated in [the documentation of `postcond`](https://github.com/jmid/multicoretests/blob/2e922e4b62492616841077714fd81246e98a1e8d/lib/STM.ml#L96), but maybe we could make it extra clear somehow? (Actually, isn't the name `postcond` a tad confusing?)

In fact, is there a particular reason why `postcond` cannot take both the pre-command and post-command state as arguments? If we allowed it, we could let the users express more things in their postconditions.

jmid:
Attempt to fix #102

### DIFF
--- a/lib/STM_spec.mli
+++ b/lib/STM_spec.mli
@@ -120,5 +120,7 @@ sig
   (** [postcond c s res] checks whether [res] arising from interpreting the
       command [c] over the system under test with [run] agrees with the
       model's result. A [postcond] function should be a pure.
-      Note: [s] is in this case the model's state prior to command execution. *)
+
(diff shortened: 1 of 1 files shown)


jmid:
CI failures look familiar, so I'll merge.

PR #195, opened 2022-11-15
CI: Set caching based on the actual upstream compiler commit

shym:
The opam cache set up by `setup-ocaml` assumes that the compiler version is enough to ensure that the proper compiler is in use, but this assumption is not true for `trunk` compilers. So this adds an optional key to the common workflow so that the `cache-prefix` key is either set to v1 or to the hash of the current upstream compiler.

This also adds the benefit of logging the current commit id of the compiler used by that CI run.

Note that this PR has no impact on the Windows-CI PR, since that one always builds the OCaml compiler from sources.

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -11,6 +11,10 @@ on:
         description: 'Compiler to use'
         type: string
         default: 'ocaml-base-compiler.5.0.0~beta1'
+      compiler_branch:
+        description: 'Source branch of the compiler, to set up caching properly (must be set if CI is not using a tagged release of OCaml)'
+        type: string
+        default: ''
       timeout:
         description: 'Timeout'
         type: number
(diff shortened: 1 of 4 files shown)


jmid:
Interestingly, the `5.0.0+trunk` bytecode run - which should be the latest `trunk` - also exhibits a `domain_spawntree` failure that we have previously observed in #179:
https://github.com/jmid/multicoretests/actions/runs/3471880569/jobs/5801960236
```
random seed: 357883174

[ ]    0    0    0    0 /  100     0.0s domain_spawntree - with Atomic
File "src/domain/dune", line 30, characters 0-114:
30 | (rule
31 |  (alias runtest)
32 |  (deps domain_spawntree.exe)
33 |  (package multicoretests)
34 |  (action (run ./%{deps} --verbose)))
(cd _build/default/src/domain && ./domain_spawntree.exe --verbose)
Command got signal SEGV.
[ ]    0    0    0    0 /  100     0.0s domain_spawntree - with Atomic (generating)
```

PR #197, opened 2022-11-16
Disable ref thread tests in bytecode mode

### LINKED ISSUE #186: Disable STM ref Thread tests under bytecode?
The STM ref Thread tests are failing - almost consistently under bytecode.
They furthermore take a long time - which may cause the CI to time out.
One example (of several) is https://github.com/jmid/multicoretests/actions/runs/3448181823/jobs/5754954999
```
random seed: 441402413

[ ]    0    0    0    0 / 1000     0.0s STM int ref test with Thread
[ ]    0    0    0    0 / 1000     0.0s STM int ref test with Thread (generating)
[ ]   22    0    0   22 / 1000    74.1s STM int ref test with Thread
...
[ ]  979    0    0  979 / 1000  4841.6s STM int ref test with Thread
[ ]  982    0    0  982 / 1000  4906.8s STM int ref test with Thread
[✓] 1000    0    0 1000 / 1000  4915.3s STM int ref test with Thread

[ ]    0    0    0    0 / 1000     0.0s STM int64 ref test with Thread
[ ]   48    0    0   48 / 1000    57.9s STM int64 ref test with Thread
[ ]  103    0    0  103 / 1000   127.7s STM int64 ref test with Thread
...
[ ]  939    0    0  939 / 1000  1413.7s STM int64 ref test with Thread
[ ]  973    0    0  973 / 1000  1474.6s STM int64 ref test with Thread
[✗] 1000    0    0 1000 / 1000  1515.2s STM int64 ref test with Thread
```
4915 secs corresponds to 81 mins or 1h 20mins and 1515 secs corresponds to 25 mins.
There's thus 1h45mins to be gained in this run.

Overall, the failures will either force us to scan CI logs for 'usual suspects' - or worse: to start ignoring CI failures.

Before we disable it, we should make sure that the above is caused by, e.g., a long interleaving prefix check in the generator rather than a Thread performance regression in OCaml bytecode mode (which needs reporting upstream).

jmid:
The goal of this PR is to fix #186 

- When looking at, it was hard to spot the right test names in the `src/neg_tests`. The first commit therefore normalizes file names to a common prefix `{stm_tests,lin_tests,lin_tests_dsl}`, followed by the mode `{sequential,domain,thread,effect}`.
- The second commit splits `lin_tests_thread.ml` into two: one for `ref` and `conclist`
- The 3rd and 4th commits handles the disabling
- The 5th commit reduces the test count of the `STM int ref Thread` test. Locally on my machine this reduces the runtime from ~14min (877secs) to ~3min (195secs) for a test we don't expect to fail anyway.

I've tested this locally on `5.0.0~beta1` and  `5.0.0~beta1+bytecode`


jmid:
I'll merge this. CI had
- a well-known `Buffer` issue triggered on `5.0.0+trunk`
- a puzzling `Ephemeron` segfault on `5.0.0+trunk bytecode`

Overall:
- Bytecode runs took 174m-248m - so no timeouts, yay!
- Native runs took 41m-76m

PR #198, opened 2022-11-16
CI: Limit the parallelism on macOS runners

shym:
We see many runs on macOS failing on SIGKILL during the compilation of dependencies, which might be a sign that their build might exceed the ressources of the runner, so we remove parallelism during that phase.

The builds went ok on pushing to my fork. Let see if that works on the builds for the PR.


jmid:
Merging as both the native MacOSX and Linux workflows run fine :+1:

PR #199, opened 2022-11-16
CI configuration for running a single test

shym:
Setting up a full CI run for a single test is expensive: we should make the most of it by running it a couple of times, either to check for the reproducibility of a given seed or to look for seeds that produce failures.
Merging all the steps also makes logs a bit cleaner for normal runs, where only that step will be skipped.


jmid:
I'll merge this.
The two bytecode failures is from well-known `STM ref Thread` tests.

Issue #168, opened 2022-10-26
Add package introduction to documentation


jmid:
The documentation for `Lin_api` and `STM` is shaping up, however it currently consists of only an enumeration of the public interface, without a high-level description of how to use it.

For a new user it would be nice to include a documentation introduction, perhaps along the lines of QCheck: https://c-cube.github.io/qcheck/0.19/qcheck-core/QCheck/index.html

jmid:
Fixed by #200 #221 and #222

PR #200, opened 2022-11-17
Stm documentation

n-osborne:
This PR propose some documentation front page for `qcheck-stm` (related to #168).

The basic idea is to have a running example of a mutable list with constant time `length` operation (a record with two mutable fields, one for the content and another one keeping track of the length).

In order to demonstrate different aspects and features of `qcheck-stm`, multiple versions of the example are proposed:

- a sequentially buggy one with sequential tests (forgot to update the `length` field in the `add` function
- a fix to this version  with successful sequential tests
- parallel test of the previous version to show that it is not parallel safe (parallel read and writes on mutable data)
- a very inelegant fix (mutex) to show the successful tests with Domains
- adding a `remove` feature to demonstrate the possibility to tune the program generator according to the state of the model (/!\ according to a sequential run) and how to use the `state` parameter in `postcond`.

### DIFF
--- a/doc/stm/index.mld
+++ b/doc/stm/index.mld
@@ -2,7 +2,7 @@
 
 {1 Content}
 
-- {!module-STM}.
+- {!module-STM} is a base module for specifying model-based state-machine tests.
 - {!module-STM_sequential} exposes a functor that allows to test a library sequentially.
 - {!module-STM_domain} exposes a functor that allows to test a library in parallel (with domains).
 - {!module-STM_thread} exposes a functor that allows to test a library in concurrency (with threads).


jmid:
I've only had a brief skim - but what I've seen looks great! :smiley: 

It seems the `index.mld` uses a mixture of `mem` and `elem/Elem` for membership querying.
I propose to use `mem` and `Mem` consistently in `index.mld` and in the code, as this is a function name familiar to OCaml programmers from `List`, `Hashtbl`, ...

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+    | Length -> Res (int, Lib.length sut)
+
+  let precond _ _ = true
+
+  let postcond cmd state res =
+    match cmd, res with
+    | Mem i,  Res ((Bool,_), b) -> b = State.mem i state
+    | Length, Res ((Int,_), l)  -> l = State.cardinal state
+    | Add _,  Res ((Unit,_),_)  -> true
+    | _                           -> false
```
jmid, line 124: Indentation is a bit off here

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+end
+]}
+
+In [arb_cmd]'s definition, note that [QCheck.make] takes an optional printer.
+It is important to provide it so that the test's output is readable. Here, we
```
jmid, line 129: `is readable` -> `is printed`?

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+use a [ppx_deriver], but you can provide your own printers if you want to.
+
+Another remark worth making, is that the [state] argument in the [postcond]
+function is the state of the model {i before} running the command (that is
+before computing [next_state cmd state]). This will be important when we will
+add a {{!section-v5}[remove]} function to our set library.
+
+Now, we are set to run our first [qcheck-stm] tests!
```
jmid, line 137: I think it is worth taking the reader by the hand here through the above definitions, ideally explaining how points 1,2,3 above are expressed with them.

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+
+We will begin with some sequential testing. They are declared like that:
+
+{[
+module Lib_sequential = STM_sequential.Make(Lib_spec)
+
+let _ = QCheck_runner.run_tests ~verbose:true [Lib_sequential.agree_test ~count:100 ~name:"STM sequential tests"]
```
jmid, line 144: This is colored strangely by `odoc`.
Does it help to add a newline before `[Lib_sequential...]`?

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+
+In the case of a failing test, [qcheck-stm] gives the user the counter example
+it has found, after some shrinking steps, as it is customary in property based
+testing {i à la} QuickCheck. The counter example is a program. It is given
+twice. The first time just as a list of calls. The second time, each call is
+paired with the the result of the computation.
+
+Looking at the output, we see that [Add 3591134320860609976] returns a [unit],
+which is to be expected. But the call to [Length] returns [0] despite the fact
+that it has been ran after the [Add]. We conclude that there is something wrong
```
jmid, line 192: has been *run*

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+      t.content <- a :: t.content;
+      t.length  <- t.length + 1;
+    end
+
+  ...
+
+end
+]}
+
+Then, the output looks like that:
```
jmid, line 217: When rerunning the sequential tests, the output now looks like that:

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+success (ran 1 tests)
+]}
+
+We can see that the tests are successful. That means that the real behaviour of
+the one hundred generated programs were consistent with the functional model
+we've given.
+
+{2 Third version -- parallel tests}
+
+We've now tested that, when run sequentially, our library behave accordingly to
```
jmid, line 234: *behaves*

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+Let's say we want to test our library when it is used with OCaml domains. We
+just have to instantiate our specification with another functor:
+
+
+{[
+(* same as before *)
+
+module Lib_domain = STM_domain.Make(Lib_spec)
+
+let _ = QCheck_runner.run_tests ~verbose:true [Lib_domain.agree_test_par ~count:100 ~name:"STM parallel tests"]
```
jmid, line 248: Again funny syntax highlighting. Add newline?

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+]}
+
+In the dune file, we should also update the dependency: [qcheck-stm.sequential]
+and [qcheck-stm.domain] are two different libraries. In the same manner, if you
+want to test your library with [Thread]s, you'll depends on
```
jmid, line 253: `you'll depends on` -> `you will need a dependency on`

I would perhaps prefer not to advertise the `Thread`-mode in the tutorial, since it is still preliminary and instead focus on the `Domain` mode that works well.

stm-doc/index.mld
```diff
@@ -0,0 +1,451 @@
...
+[✓]  100    0    0  100 /  100    37.8s STM parallel tests
+================================================================================
+success (ran 1 tests)
+]}
+
+This time, the test succeed!
+
+{2 Fifth version -- adding a [remove] function}
+
+Now that our implementation is domain safe, let's add a new feature: a [remove]
```
jmid, line 380: `safe for parallel usage`?

jmid (commented):
Thanks again. I've now taken a pass over the text and commented on a number of smaller things.

Overall, I propose we integrate the `{1 Content}` section at the end with the summary at the top, so that users don't have to scroll through the tutorial when they want to look up in the library documentation.

Taking a step backward, the tutorial implements a set with a `list` and tests it against a better model set implementation using a balanced search tree, that is more complex, see, e.g., to this `add` function from `set.ml` in `Stdlib`:
```ocaml
    let rec add x = function
        Empty -> Node{l=Empty; v=x; r=Empty; h=1}
      | Node{l; v; r} as t ->
          let c = Ord.compare x v in
          if c = 0 then t else
          if c < 0 then
            let ll = add x l in
            if l == ll then t else bal ll v r
          else
            let rr = add x r in
            if r == rr then t else bal l v rr
```
This muddles the water between 'simple model' and 'more complex sut code'.
I think it would be more instructive to switch roles:
- use a `mutable Set` in `Lib` (which has a O(n) `cardinal` function and so would benefit from the O(1) improvement)
- test it against a simple `list`-based model

stm-doc/index.mld
n-osborne: Finally, I've removed the dune file from the tutorial (it's mostly noise). And I agree about not advertising the `Thread`-mode in the tutorial.

stm-doc/index.mld
n-osborne: It is Indeed strange looking syntax highlighting. I've tried several gymnastic but none have changed the output.

jmid: Fair enough, thanks for investigating!
Perhaps `list` literals are just colored differently? :thinking:

doc/stm/index.mld
```diff
@@ -0,0 +1,581 @@
...
+  mode is currently experimental as it does not trigger as many bugs as the
+  [domain] mode. This is due to a difference in their execution models: domains
+  can run in parallel (with all the races that it can induce) while threads are
+  interleaved so that races require specific scheduling to be revealed.
+
+{1 Example: how to test a library?}
+
+Suppose we want to implement a small mutable set library, our main focus being
+to have a constant time [cardinal] operation. We will be using [Stdlib.Set] for
+the content, keeping track of the cardinal when adding and removing elements.
```
jmid, line 37: nit: *of the cardinality` 
https://en.wikipedia.org/wiki/Cardinality

doc/stm/index.mld
```diff
@@ -0,0 +1,581 @@
...
+and [sut] and wraps the result with information about its type in a
+{{!STM_base.res}res} type.
+
+{3 The model definition}
+
+[state] is the type of the model. The model should be pure and simple enough to
+be obviously correct. Here we use a simple {{!Stdlib.List}list} from the
+standard library.
+
+[qcheck-stm] also needs to know what is the initial value. This is provided by
```
jmid, line 179: Suggestion:
the model's initial value

doc/stm/index.mld
```diff
@@ -0,0 +1,581 @@
...
+function will come in handy when we will add a {{!section-remove}[remove]}
+function to our set library.
+
+{3 The [cmd]s generation}
+
+The last thing to do is to tell [qcheck-stm] how to generate the [cmd]s. This
+is what is done by the [arb_cmd] function, using [QCheck]'s combinators. We
+will make two remarks on this function.  The first one is that [QCheck.make]
+takes an optional printer. It is important to provide it so that the test's
+output is readable.  Here we use the function [show_cmd] that has been built by
```
jmid, line 203: 'readable' may sound as if it is otherwise unreadable in the sense of being "garbled".
If that is the case, we should open an issue and fix it.
It not, I think that writing *"so that the test's output is printed"* is clearer.

doc/stm/index.mld
```diff
@@ -0,0 +1,581 @@
...
+   Cardinal : 0
+================================================================================
+failure (1 tests failed, 0 tests errored, ran 1 tests)
+]}
+
+In the case of a failing test, [qcheck-stm] gives the user the counter example
+it has found, after some shrinking steps, as it is customary in property-based
+testing {i à la} QuickCheck. The counter example is a program. It is given
+twice: the first time just as a list of calls; the second time, each call is
+paired with the result of the computation.
```
jmid, line 261: Should we fix `STM` so that the printing is more consistent as `cmd`s under each other (no `;` or square braces)?

doc/stm/index.mld
```diff
@@ -0,0 +1,581 @@
...
+
+We can see that the output is a bit different than the one with the failing
+sequential test. This is because the generated programs are a bit different
+too. For testing with using [Domain], [qcheck-stm] generates a triplet of list
+of calls: one sequential prefix to bring the system under test in a random
+state and two parallel suffixes.
+
+We can see in the minimal counter example proposed by [qcheck-stm] that there
+are some parallel writes and reads occurring. This is a classical mistake when
+writing parallel code known as data race.
```
jmid, line 379: as *a* data race

doc/stm/index.mld
```diff
@@ -0,0 +1,581 @@
...
+
+Let's fix that by putting some lock on every operation. This will end up being
```
jmid, line 381: putting *a* lock

doc/stm/index.mld
```diff
@@ -0,0 +1,581 @@
...
+      match state with
+      | [] -> Gen.int
+      | xs -> Gen.(oneof [oneofl xs; int])
+    in
+    QCheck.make ~print:show_cmd
+      (QCheck.Gen.oneof
+        [Gen.return Cardinal;
+         Gen.map (fun i -> Mem i) Gen.int;
+         Gen.map (fun i -> Add i) Gen.int;
+         Gen.map (fun i -> Remove i) gen;
```
jmid, line 544: By the same argument
- `Mem` will rarely ask for an actual member
- `Add` will rarely test for adding an existing member

For this reason I think they should also be updated to use the `gen` generator

doc/stm/index.mld
```diff
@@ -0,0 +1,581 @@
...
+================================================================================
+failure (1 tests failed, 0 tests errored, ran 1 tests)
+]}
+
+We see in the output that there is indeed a [Remove] of a previously added
+element. So this execution path is
+now explored. As there is only one [Add], one [Remove] of the added element and
+then [Cardinal] returns [1], we can easily conclude there is something wrong in
+the implementation of the removal of an element when the element is indeed in
+the set.
```
jmid, line 581: I would have expected the `Mem` call to have been shrunk away in the minimal counterexample? :thinking: 

I suggest choosing another counterexample and filing an issue about this deterministic, sequential example not shrinking to the minimal one.

jmid (commented):
OK, I've now taken a second pass over this and commented on a few smaller things.

At the end, one could add a section about adding an optional `cmd` shrinker, but we can always do so in a separate PR.

doc/stm/index.mld
n-osborne: Yes, I agree.
And it is the occasion to add a small remark to draw the reader's attention on the fact the problem is not always obvious :-)

doc/stm/index.mld
n-osborne: Yes, I think it would be nicer.

doc/stm/index.mld
```diff
@@ -0,0 +1,583 @@
...
+         Gen.map (fun i -> Add i) gen;
+         Gen.map (fun i -> Remove i) gen;
+        ])
+]}
+
+This will be enough to trigger a failure in the test which can help us correct
+our implementation.
+
+{[
+random seed: 284906494
```
jmid, line 554: The recent commit c598032 introduced a duplicate seed.
This one can be removed.

doc/stm/index.mld
jmid: I see you've created #218 for this and have a PR coming :+1:

doc/stm/index.mld
jmid: I know of two approaches to help notice it:
- *fault injection*, e.g., if we alter `mem` to always return `false` we should wonder why tests still succeed with that change
- printing samples from the `cmd` generators distribution with `Gen.generate ~n:10` or `Gen.generate1`

It may be good to document these tricks at some point - but I think it is fair enough to leave them out for this first version.

jmid (commented):
Thanks for the prompt fixes!
I spotted a duplicate `seed` being introduced, but the rest I think we can handle in follow-up PRs :+1:

jmid:
Dang, I don't know why GitHub marked this as closed - sorry about that! :grimacing: 
To make it clear: this has been merged!
I just checked it out and rebased locally to update it to the `Base`-renaming and check that things compiled afterwards.

PR #204, opened 2022-11-18
Remove kcas tests

jmid:
This PR removes the old `kcas` tests.

If we ever find ourselves in a situation where we want to revive them, they are available on a dedicated `add-kcas-tests` branch: https://github.com/jmid/multicoretests/tree/add-kcas-tests


jmid:
Merging this. CI logs show
- known `5.0.0~beta1` bytecode crashes on `Ephemeron` and `Effect`
- a likely deadlock on `threadomain`

PR #207, opened 2022-11-22
Lin renames

shym:
Following, and as planned in, #164, this PR proposes a few changes of names, now that the internals of Lin are hidden away.


jmid:
I experimented with inlining the extra modules in `Lin` and `STM` yesterday.
As part of that I initially renamed `CmdSpec` and then ended up reverting the renaming.
To open the discussion I'll create PRs to share the inlining attempts.

PR #208, opened 2022-11-22
More post-refactor Lin polishing

jmid:
This PR
- simplifies the `{Lin_domain,Lin_thread,Lin_effect}.Make` type annotation
- adds minimal interfaces for  `Lin_domain`, `Lin_thread`, and `Lin_effect`
- updates the doc/paper-examples/lin_tests_dsl code

PR #209, opened 2022-11-22
Inline Lin modules

jmid:
This PR builds on top of #208 so the first 5 commits can be ignored.

With the PR's inlining, we end up in a situation
- with 4 Lin `ml`+`mli` files for `Lin_base`, `Lin_domain`, `Lin_thread`, and `Lin_effect` - so one pair for each of the 4 user-facing modules
- where the old raw/pure tests can just `open Lin_base.Internal`

To do so
- it creates a `Lin_base` interface
- renames `Lin_internal` to `Lin_base.Internal`
- eliminates `Lin_common` (which was `include`d anyway)


jmid:
Merging this after Slack consensus...

PR #210, opened 2022-11-22
Inline STM modules

jmid:
This PR supplements #209 by performing a similar module inlining for `STM`.

It inlines the `STM_internal` and `STM_spec` modules in `STM_base`, leaving us with 4 `ml`/`mli` file pairs for each of the user-facing modules: `STM_base`, `STM_sequential`, `STM_domain`, and `STM_thread`.

For now I went with the `STM_spec` name for the module type, even though when it lives in `STM_base` `STM_base.STM_spec` becomes a bit redundant. We might want to go with the more eye-friendly module path `STM_base.Spec`... :thinking:


jmid:
Merging this after Slack consensus...

Issue #192, opened 2022-11-15
Add alerts on internal and experimental modules


jmid:
Before a release we should add `@@@alerts` on modules that are either
- internal, such as `STM_internal` or
- experimental, such as `STM_thread`

PR #211, opened 2022-11-22
Alerts internal

shym:
To address half of #192, set up alerts on uses of the `internal` modules and explicitly disable them on expected sites in `lib`.
This PR is split in various commits so that the alerts do show up before being disabled.


jmid:
Thanks, this looks good! :+1: 
Do we also want to add internal warnings to `Lin_domain.Make_internal`, `Lin_thread.Make_internal`, ...?

shym:
We certainly should, so here we go!

jmid:
Thanks!

PR #223, opened 2022-11-25
Alert experimental

n-osborne:
To address the other half of #192 , the sister PR of #211 
Set up an alert on uses of `STM_thread.Make` and explicitly disable it on uses in `src/neg_tests`.

### DIFF
--- a/lib/STM_thread.mli
+++ b/lib/STM_thread.mli
@@ -19,4 +19,4 @@ module Make : functor (Spec : STM.Spec) ->
         [count] is the test count and [name] is the printed test name. *)
 
   end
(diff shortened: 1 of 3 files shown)


jmid:
I just pushed commit 4aff086 to elaborate a bit on the warning description

jmid:
Dang, the same thing happened again :grimacing: 
There were merge conflicts with this PR because 15d850c and 04df79b were also adding `-experimental` warnings on `Lin_thread.Make_internal` and `Lin_effect.Make_internal` covered by #211. I therefore locally made commits to undo these changes, squashed these into the original commit, and merged. Oh well.

PR #226, opened 2022-11-28
CI: Ensure at most one simultaneous run per PR

### LINKED ISSUE #225: Cancel a PR's earlier CI jobs on subsequent pushes
We sometimes find ourselves in a situation where
- you open a PR, kicking off a long CI-run
- you spot an oversight in your PR and push a commit fix, thereby launching a second, long CI-run

On the fix-push it would be nice to automatically cancel the earlier CI jobs.
(for context, I saw a recent PR on ocaml/ocaml providing this)

shym:
Updating a PR will cancel all the current runs
We still keep the runs on standard push, by using `head_ref` which is defined only on PR events or falling back on `run_id` which is unique otherwise

Closes #225.

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -1,5 +1,9 @@
 name: Common CI workflow
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 on:
   workflow_call:
     inputs:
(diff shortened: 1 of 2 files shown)


shym:
I intentionally only did this for `common` first, in order to force an update of the PR with `windows` covered, and then check that all `common`-based workflows were indeed cancelled.

jmid:
Thanks, this is great!
Love the meta-level testing :heart: :smile: 
I'll let the CI finish and then merge.

PR #227, opened 2022-11-28
Simplify src/internal tests

jmid:
This PR simplifies the `src/internal` tests a bit by using only one 'stanza' (rule) per test.

It furthermore uses `dune`'s built-in support for `diff`ing with `.expected`-files, thus simplifying the output which sometimes clobbers the output in CI-logs...


jmid:
Merging after having confirmed that this indeed clears up the CI-logs a bit more :tada:

PR #229, opened 2022-11-29
Some CI improvements

shym:
In particular bring help to trigger one-shot CI runs to torture one particular test easily.

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -79,7 +79,7 @@ jobs:
           cache-prefix: ${{ steps.compute_prefix.outputs.cache_prefix }}
 
       - name: Override compiler to a particular commit
-        if: inputs.compiler_commit != ''
+        if: env.OCAML_COMPILER_COMMIT != ''
         run: |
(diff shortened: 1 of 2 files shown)


jmid:
Thanks, LGTM!

I'm wondering about `inputs.compiler` (in both windows.yml and common.yml) since it is one of the few `inputs.` variables remaining the workflow body. Can we, e.g., be in an overriding situation where the wrong title is printed, but the right version is run in this printed line in windows.yml?
https://github.com/jmid/multicoretests/blob/44fb3d6f36cbc7c244805263265196c1f903f1f1/.github/workflows/windows.yml#L50

shym:
Good remark!
So that makes me notice that the (input) keys to choose the actual compiler to test have slightly different meanings in the `common` and the `windows` workflows. That’s annoying, but due to the fact that they are installed in completely different ways, I’m not sure how to reconcile them at the moment.

jmid:
I noticed that too. I think it is fair enough to keep as is for this reason :+1:

PR #236, opened 2022-11-30
Less 'linearizability'

### LINKED ISSUE #232: Tone down occurrences of 'Linearizability'
The term 'Linearizability' has a formal meaning within concurrency.
What we test with `Lin` and `STM` is closer to the term 'Sequential consistency', hence it would make sense to tone down occurrences of the former.

For context:

In the initial parallel QuickCheck paper "Finding Race Conditions in Erlang with QuickCheck and PULSE"
https://publications.lib.chalmers.se/records/fulltext/125252/local_125252.pdf
Claessen et al. write (Sec.4.1):

> In order to test for race conditions, we need to generate test cases
> that are executed in parallel, and we also need a specification of the
> correct parallel behavior. We have chosen, in this paper, to use a
> specification that just says that the API operations we are testing
> should behave atomically.
>    How can we tell from test results whether or not each operation
> “behaved atomically”? Following Lamport (1979) and Herlihy and
> Wing (1987), we consider a test to have passed if the observed
> results are the same as some possible sequential execution of the
> operations in the test—that is, a possible interleaving of the parallel
> processes in the test.
>     Of course, testing for atomic behavior is just a special case,
> and in general we may need to test other properties of concurrent
> code too—but we believe that this is a very important special case.
> Indeed, Herlihy and Wing claim that their notion of linearizability
> “focuses exclusively on a subset of concurrent computations that
> we believe to be the most interesting and useful”; we agree. In
> particular, atomicity is of great interest for the process registry.
>    One great advantage of this approach is that we can reuse the
> same specification of the sequential behavior of an API, to test its
> behavior when invocations take place in parallel. We need only find
> the right linearization of the API calls in the test, and then use the
> sequential specification to determine whether or not the test has
> passed. We have implemented this idea in a new QuickCheck module,
> eqc_par_statem, which takes the same state-machine specifications as
> eqc_statem, but tests the API in parallel instead. [...]

On this basis, I think it should be fair enough to use 'linearization' and 'to linearize'.
The term 'sequential interleaving' is also representative.

jmid:
This PR fixes #232

This rephrases occurrences of the term *'linearizability'* which has a specific meaning within concurrency https://en.wikipedia.org/wiki/Linearizability and therefore may confuse more than it will help.

There are updates
- in dune files
- in .ml and .mli files
- in markdown and .mld files

While changing occurrences in the LaTeX/Markdown paper source, I also updated it to the new Lin+STM interfaces,
as there is no point in sharing broken code relying on the old interface.


jmid:
CI failure summary:
- a MacOSX failure to trigger the Ephemeron unsafety in 1000 iterations
- a Windows `threadomain` crash
- a Windows linker error

Merging

PR #237, opened 2022-11-30
Two little CI tweaks

shym:
Two little tweaks for the `common`-based workflows:
- add a counter when we repeat a test
- add the option to stop repeating that test at the first failure.


jmid:
I think this is a good idea :+1: 
I haven't used this CI feature much myself - but comparing to the equivalent `Util.repeat` functionality, I was wondering whether we should make 'stop-early' the default?

jmid:
I'll merge this. We can always change the default later.

PR #238, opened 2022-12-01
A few corrections and suggestions for Lin documentation

shym:
This PR creates one link for `option` but it’d be nice also to get the links to `QCheck` values working too.


jmid:
Wonderful, thanks for spotting these! :pray: 
I agree about adding more links with `{!...}` - but we can do so in separate PRs.

I'll let the CI finish and then merge :+1:

PR #240, opened 2022-12-02
Interface revision

jmid:
This PR takes a run over the interfaces.

In `STM` and the `Lin_internal` functors it strives to expose a pair (generator and property)
- with consistent suffixes (currently `_par` or `_conc` - but I am not insisting on these names)
- which allow a client to write a custom `lin_test` to play with `seq_len` `par_len` `rep_count` and `retries`

Exposing only one of these two is insufficient to do so, as a QCheck test requires both.
To achieve it requires a bit of renaming and exposing a function/alias here and there.

Finally the PR elaborates on `Util.repeat` stopping early on the first failure.


lib/STM_thread.ml
```diff
@@ -9,6 +9,8 @@ module Make (Spec: Spec) = struct
 
   exception ThreadNotFinished
 
+  let arb_cmds_conc = arb_cmds_par
```
shym, line 12: This reads a bit strange, maybe the generic generator `arb_cmds_triple` now that `_par` and `_conc` are used consistently?
(This is arguably only quibbling ;-)

lib/STM_thread.mli
```diff
@@ -4,6 +4,11 @@ module Make : functor (Spec : STM.Spec) ->
   sig
     exception ThreadNotFinished
 
+    val arb_cmds_conc : int -> int -> (Spec.cmd list * Spec.cmd list * Spec.cmd list) QCheck.arbitrary
+    (** [arb_cmds_par seq_len par_len] generates a [cmd] triple with at most [seq_len]
```
shym, line 8: -> arb_cmds_conc

lib/STM_thread.mli
```diff
@@ -4,6 +4,11 @@ module Make : functor (Spec : STM.Spec) ->
...
+        sequential commands and at most [par_len] parallel commands each.
```
shym, line 9: Should we reformulate _parallel_ here?

shym (approved):
Nice clean up!

lib/STM_thread.ml
jmid: Indeed - addressed by renaming to the universal `_triple` suffix

lib/STM_thread.mli
jmid: Well spotted!

PR #241, opened 2022-12-02
Add some links in documentation

shym:
Still missing: passes over Lin tutorial and double-checking that links are actually working (using `odig doc` on locally installed packages, or is there a simpler way?).


jmid:
Great, many thanks for adding these! :pray: 
I'm not much into `odoc`. I tend to just generate and open the documentation locally and then click the relevant link :shrug:

PR #247, opened 2022-12-06
Doc more detail review

shym:
My review of #245. It could all be squashed into the original PR (maybe 1 commit for STM, 1 for Lin?).


jmid:
Great, thanks!

PR #248, opened 2022-12-06
Add `returning_ t` and constants in Lin DSL

shym:
The last commit of this series is really there to show what the other two add, but should probably not be included: it is probably wasting time on useless test cases.


shym:
The PR is now beheaded.

PR #249, opened 2022-12-07
add missing (package ...) and (public_name ...) annotations

jmid:
We are missing package annotations on some of the `dune` rules - which means, e.g., a qcheck-stm installation will try to build src/domainslib executables without the dependencies present.

Adding `(package multicoretests)` on everything is not enough.
For `(executable ...)` stanzas they should be declared with a `(public_name ...)` too.
The PR therefore updates `(name ...)` to `(public_name ...)` on the relevant `(executable ...)` stanzas.

### DIFF
--- a/src/array/dune
+++ b/src/array/dune
@@ -10,7 +10,8 @@
    lin_tests_dsl.exe))
 
 (executable
- (name stm_tests)
+ (public_name stm_tests)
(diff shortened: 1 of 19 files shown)


jmid:
Closing this in favor of #250

PR #250, opened 2022-12-07
Simplify dune rules into (test ...)s and add package annotations

jmid:
This PR offers an alternative solution to the missing `(package ...)` annotations and the solution suggested in #249.
Instead it merges `(executable ...)` and `(rule ...)` stanzas into `(test ...)s`
while adding the missing `(package …)` annotations.
This has the advantage of simplifying our dune rules.


jmid:
Merging this as I believe the bytecode and Windows install failures are unrelated

PR #253, opened 2022-12-08
CI: Switch to OCaml 5.0 RC 1

shym:


jmid:
Wonderful, thanks! :pray:

PR #254, opened 2022-12-08
Update a few links in the README

shym:
Still leaves a couple of reference to the `jmid/multicoretests` in doc: should those also be updated?


jmid:
Well spotted, thanks!

PR #255, opened 2022-12-08
Add the debug runtime in dune config and CI

shym:


jmid:
This is great! :pray:

I was trying this out myself yesterday and from `man dune build` I learned that `dune` will fall back on the `DUNE_PROFILE` environment variable. If those work in CI-mode too, that means we could avoid some of the `--profile=...` logic, no? I'm basically clueless about GitHub Action workflow programming and therefore prefer to keep these as simple as possible... :sweat_smile: 

We should also add a badge! (better rebase first because the rc1 changes at the top of README will cause conflicts)

jmid:
The OCaml CI runs this with [extra options](https://github.com/ocaml/ocaml/blob/trunk/.github/workflows/build.yml#L193-L195) for [heap verification](https://github.com/ocaml-multicore/ocaml-multicore/wiki/Debugging-the-OCaml-Multicore-runtime#heap-verification-and-debug-mode)

Should we perhaps do so too? :thinking:

shym:
When in CI, do as the OCaml developers do!
I love the cleanup of the `dune` invocations coming from using the environment variable directly. Note that I had picked up the correct name already :smiley:
I then realised that `dune build` would fail on `dev` profile (I thought it was the default), hence #257, which is reused in that PR.
(I admit, while I was working on `env`, I’ve used it to set `QCHECK_MSG_INTERVAL`, to drop that one extra step from the logs).

jmid:
Rebased on `main` and dropped the commit 61b0c3c which conflicted with #256.
Also add a `trunk-debug` workflow and badges.

jmid:
Merging this. We can handle the overflowing badges in #264.

PR #256, opened 2022-12-09
Dune file cleanup

jmid:
Following the dune rule simplification in #250 this PR offers another benefit:
We get rid of the pesky `"this prevents the tests from running on a default build"`-rules!
These were again a historical artifact of the early, hand-written test `(rule ...)`s - which we didn't want triggered on a `dune build`. Since we don't use those anymore  - we don't need the extra prevention rule at the top anymore.
Overall #250 and this PR simplifies our dune rules considerably :tada: 

I left the single remaining rule in the top-level `dune` file for now (until I recall why I added it... :sweat_smile:)

The clean-up triggered a bunch of warnings/errors about unused definitions, so the following commits address these.


src/buffer/stm_tests.ml
```diff
@@ -25,7 +25,7 @@ struct
   type state = char list (* in reverse *)
   type sut = Buffer.t
 
-  let shrink_cmd c = match c with
+  let _shrink_cmd c = match c with
```
shym, line 28: In the other commits of that PR you removed the unused code.
You want to keep this one in case it turns out useful later?

shym (approved):
Nice clean-up!

src/buffer/stm_tests.ml
jmid: Yep, exactly :+1:

PR #259, opened 2022-12-09
Pin lockfree to the last version before dscheck dependency

### LINKED ISSUE #258: Build of dependencies failing consistently on bytecode target
Since Wednesday the 7th, between 16:00 and 17:00 UTC, all builds on bytecode started to fail.
I traced it to ocaml-multicore/lockfree#34 which pulls in (recursively) many new dependencies, some of which cannot build on bytecode (`containers`).

shym:
Some recursive dependency of dscheck does not build on bytecode-only OCaml so pin lockfree before that change to keep that target.

Hopefully should close #258.

### DIFF
--- a/multicoretests.opam
+++ b/multicoretests.opam
@@ -49,5 +49,5 @@ build: [
 ]
 dev-repo: "git+https://github.com/ocaml-multicore/multicoretests.git"
 pin-depends: [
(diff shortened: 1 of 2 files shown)


jmid:
"All checks have passed" - very happy to see this again :smile: Good detective work, sir!

PR #263, opened 2022-12-12
Cleanup resources after each domain-based Lin test

shym:
In domain-based test, `sut` was never cleaned-up, and so was eventually leaking resources

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## Not released yet
+
+- #263: Cleanup resources after each domain-based Lin test
+
 ## 0.1
 
 The initial opam release of `qcheck-lin`, `qcheck-stm`, and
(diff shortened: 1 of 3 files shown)


jmid:
We should add a `CHANGES` entry - and aim for a bug-fix release.

Also, since this is central functionality I was hoping we could add a test case for it in `src/internal` to avoid it ever happening again.

There's a certain beauty to having messed up "double clean-up" in `Lin` and having a test case for it in `src/internal/cleanup.ml` (we should perhaps rename it to `doublecleanup`) and now getting hitting by a missing `cleanup`...  :facepalm:

shym:
I thought the two tests about cleanups could merged together.
I checked locally that the test fails on `main`. But it admittedly doesn’t give a very useful feedback (just saying that the empty run is failing). As it’s an internal test, it might be good enough?

jmid:
Great, thanks!

Issue #260, opened 2022-12-12
Shorten CI badge labels


jmid:
Our increasing list of target workflows causes line overflow as can be seen on
https://github.com/shym/multicoretests/tree/profile-debug
![Screenshot from 2022-12-12 10-49-04](https://user-images.githubusercontent.com/2655248/207014476-a2a8a9ec-a4ab-449a-be6a-180463f45b09.png)

There's however quite some redundancy in the badge labels.
I'm imagining we could pull that redundancy out as "line labels":
```
5.0.0~rc1:   [Linux|passing] [MacOSX|passing] [Linux-bytecode|passing] ...
5.0.0+trunk: [Linux|passing] [MacOSX|passing] [Linux-bytecode|passing] ...
```

For some reason I haven't figured out yet, the typeset label is the `name` field from the workflow file, rather than the supplied `[label]` as described in the manual: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge. For example, the bytecode label is set as `Linux 5.0.0~rc1-bytecode` (note the last dash `-`):
https://github.com/ocaml-multicore/multicoretests/blob/e409e0e37c814ce42ad43563837a20695a45e5f8/README.md?plain=1#L4
whereas what is typeset is the `name` field (note the missing dash `-`):
https://github.com/ocaml-multicore/multicoretests/blob/e409e0e37c814ce42ad43563837a20695a45e5f8/.github/workflows/linux-500-bytecode-workflow.yml#L1

So, I think we'll need to either to figure out the problem - or bite the bullet and change the workflow `name`s :shrug:

shym:
I didn’t see in the documentation mentions of the label.
As the label is embedded in the svg you get when you request https://github.com/ocaml-multicore/multicoretests/actions/workflows/linux-500-bytecode-workflow.yml/badge.svg, it must be based on the content of the `.yml` file, mustn’t it?

jmid:
I think you are right. The first `[label]` looks like it is an "alt text" for the generated status image :shrug:

jmid:
Perhaps changing workflow `name` is a bad idea, since it will affect the overview page at https://github.com/ocaml-multicore/multicoretests/actions and hence make it harder to identify the platform and version of a failing run from it... :thinking: 

There are third party solutions it seems: https://github.com/marketplace/actions/bring-your-own-badge
Looks like we could customize the badge to our liking with it - without changing `name`?

PR #264, opened 2022-12-12
Shorten workflow names by dropping (redundant) numbers

shym:
First proposition to address #260. Names could be shortened further (Linux, Win, Mac, BC; tr for trunk) if need be.


jmid:
I just merged #255, so this now needs updating to include debug runtime

Issue #193, opened 2022-11-15
Move tests to Domainslib and Lockfree after release


jmid:
After a release we should create PRs to move
- `src/lockfree` tests to the lockfree repo https://github.com/ocaml-multicore/lockfree
- `src/domainslib` tests to the domainslib repo https://github.com/ocaml-multicore/domainslib

PR #265, opened 2022-12-12
Remove lockfree tests

jmid:
I've created a PR to move the `lockfree` `Ws_deque` test to that repo, now that we have released `qcheck-stm 0.1.0`:
ocaml-multicore/lockfree/pull/43, thus addressing the first half of #193.

This PR thus
- removes the test code
- cleans up its dependencies from `multicoretests`
- cleans up mention and pointers to the tests from README-markdown files

I've created the branch `lockfree-tests` as a name for the most recent `main` with the tests still present.


jmid:
CI is failing because `domainslib` depends on `lockfree` which then pulls in the broken, transitive `containers` dependency...

jmid:
The only failure is the well-known `threadomain` timeout on Windows so I'll merge.
Multiple runs on MacOS and Linux are taking less than 40min now with the removal of both the Domainslib and Lockfree tests and their dependency installation.

PR #271, opened 2022-12-14
Increase the timeout for the debug-runtime tests

shym:
The debug runtime is slower than the standard runtime, so that it regularly runs out of time during CI, so give it the same timeout as other slow platforms (bytecode and windows)


jmid:
Good point, thanks!

PR #275, opened 2022-12-20
Use (real) 5.1 trunk for the “trunk” workflows

shym:
This PR proposes to pin a ppxlib that supports 5.1 (while it works...) for the trunk runs, thinking it might be more interesting to test this branch than 5.0.1.
If it is to be used, this would also require an update to badges.


jmid:
Wow - great work Samuel!
I'm not sure what to wish for though :grimacing: I guess both red and green are Christmas colors... :sweat_smile:

jmid:
CI looks all green!
Testing the actual `trunk` is much more useful in the current situation whether the coming 5.1 has continuous PRs merged that would be useful to test. The cost of an "unofficially patched `ppxlib`" outweighs the benefit IMO.
With an update to the README badges I'm therefore ready to merge this :+1:

jmid:
Great, thanks!

PR #277, opened 2022-12-21
CI: Retry a couple of times when getting “null“ as cache key

shym:
In quite a few CI runs on trunk, obtaining the SHA of the tree of latest commit of the compiler fails, and so caching does something fairly unreliable. To avoid that, retry a couple of times to get that SHA1, waiting quite some time between retrials, to give github time to come back up

In the runs to try out that commit, I saw a couple of times “null”, and every time, asking a couple of times didn’t help, hence it will now the display the API answer to try and debug the issue further

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -82,10 +82,17 @@ jobs:
         shell: bash
         run: |
           if [ -n "$OCAML_COMPILER_BRANCH" ]; then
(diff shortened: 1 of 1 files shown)


jmid:
OK, not particularly pretty - but quite understandable. Do we need to do something similar for the Windows CI?

shym:
For better or worse, we are not using a cache on windows builds, so no need for windows.
(It might be time to try and reconsider whether we could manage to get caching in working order there, but that’s off-topic here ;-)

jmid:
Fair enough, I'll merge then :+1:

PR #278, opened 2022-12-21
CI: Set up OPAMJOBS=1 on macOS

shym:
It was previously done only for the whole set of dependencies. Now set it up so that it's used also for ppxlib if it is pinned


jmid:
Looks good, thanks!

PR #279, opened 2022-12-22
CI: Use git ls-remote to infer a reliable cache key

shym:
Use git ls-remote to obtain the latest commit on the followed branch, in order to avoid hitting the API-rate limit
Note that the previous strategy was using the identifier for the _tree_ rather than the _commit_, but the extra caching this allowed was probably never actually used

Follow up to #277 now that I saw that `null`s were simply due to the API-rate limit being reached.


jmid:
Nicely done, thanks!

PR #214, opened 2022-11-23
Initial tests of the Weak module

jmid:
I realized we didn't have tests of `Weak` yet, so I played last night with writing some.
Because the GC may kick in, the interface is generally not sequentially consistent (nor is it intended to be safe to use in parallel).

Nevertheless, this was producing strange, out-of-thin-air(?) values:
```
Messages for test Lin DSL Weak test with Domain:

  Results incompatible with sequential execution

                                               |                    
                                Weak.set t 9 Some (9) : Ok (())     
                                               |                    
                         .------------------------------------------.
                         |                                          |                    
            Weak.set t 9 None : Ok (())         Weak.get t 9 : Ok (Some (93946051950632))
```

PR #281, opened 2023-01-13
Escape printed strings in STM

jmid:
While writing `STM` tests of `Weak` `Hashset`s in #214 I noticed that the `string` combinator's built-in printer did not escape the printed characters - or add enclosing `"`-quotes. Here's an output example:
```

                                                                                   |                                      
                                             (Merge "q\148x\189xu\225\235\014n\178H0\234\195\237") : Ok (q�x�xu��n�H0���)
                                                                                   |                                      
                                           .------------------------------------------------------------------------------.
                                           |                                                                              |                                      
             (Find "q\148x\189xu\225\235\014n\178H0\234\195\237") : Ok (�)                (Remove "q\148x\189xu\225\235\014n\178H0\234\195\237") : Ok (())       
(Add "\208") : Ok (())
```
Notice that the `Merge`, `Find`, and `Add` arguments are nicely escaped and quoted, while the `Ok` result arguments are not. This is because the former were derived with `[@@deriving show]`.

In my terminal the resulting strings mostly show as question mark signs in a diamond (not so informative). 

Here's an example output with the fix - for contrast:
```

                                                                   |                               
                                                   (Add "\140\251\192\136") : Ok (())              
                                                                   |                               
                                   .---------------------------------------------------------------.
                                   |                                                               |                               
     (Find_opt "\140\251\192\136") : Ok (Some ("w\195Z\181\015a*"))              (Remove "\140\251\192\136") : Ok (())             
                                                                                   (Add "w\195Z\181\015a*") : Ok (())  
```
In this case it becomes clearer that `Find_opt` doesn't necessarily return a string identical to its argument when used in parallel (as we would expect).


jmid:
CI is happy so I'll merge this one-liner fix.

PR #284, opened 2023-01-18
Streamline CI

shym:
Thanks to @dra27’s [explanations](https://discuss.ocaml.org/t/pre-ann-installing-windows-ocaml-5-0-0-in-opam/11150) and opam repository, I finally figured out how to install the right version of the compiler at the `setup-ocaml` step under Windows. In particular, I understood how I can configure `.gitattributes` in my [custom opam repository](https://github.com/shym/custom-opam-repository) so that hashes work. So now I added there an opam package for OCaml 5.1 trunk for Windows, so now the CI configurations for Windows and Unices are so close we can finally close the gap and use a common workflows for both. I also moved to that opam repo the special case of `ppxlib` for 5.1, which cleans up quite a bit the workflow.
The biggest win for windows workflows is that (opam) caching will work its magic on the compiler package, so skipping building it every single time!

I also took the opportunity to add a workflow testing the bytecode VM on Windows too, as @dra27 thought we were doing already :smile:.
On the other hand, I wonder whether we should cut back a little on the number of workflows we run for each push. Maybe some workflows should run only under more restricted constraints, such pushes on `main` only?


shym:
I just updated this PR to take into account the fact that `$OPAMJOBS` is set up by `setup-ocaml` even when it is already set. So this must be reset after this step, we need to configure environment variables in two different steps.

jmid:
Much appreciated! Being able to focus a test in the CI by only changing in one workflow file is a definite improvement - on top of being able to benefit from the caching mechanism! :tada: :smiley:

jmid:
We should add a badge for Windows bytecode since you included it (alternatively pull it out and add both in a separate PR). Also, should we include Windows bytecode 5.1.0 too?

shym:
I admit that symmetry should require a winbyte 5.1! :smile:
Let’s push that to another PR.

jmid:
This worked so nicely in the PR - but just failed in the CI on Windows `trunk` during `opam` install:
https://github.com/ocaml-multicore/multicoretests/actions/runs/3959945012/jobs/6783427411
```
  [...]
  opam64/install.sh
  D:\cygwin\bin\bash.exe C:\hostedtoolcache\windows\opam\0.0.0.2\x64\opam64\install.sh --prefix /usr
  D:\a\_actions\ocaml\setup-ocaml\v2\dist\index.js:401
                          throw new Error(`Cache upload failed because file read failed with ${error.message}`);
                          ^
  
  Error: Cache upload failed because file read failed with EBADF: bad file descriptor, read
      at ReadStream.<anonymous> (D:\a\_actions\ocaml\setup-ocaml\v2\dist\index.js:401:31)
      at ReadStream.emit (node:events:390:28)
      at emitErrorNT (node:internal/streams/destroy:157:8)
      at errorOrDestroy (node:internal/streams/destroy:220:7)
      at node:internal/fs/streams:258:9
      at FSReqCallback.wrapper [as oncomplete] (node:fs:660:5)
```

shym:
Ouch! My only hope is it would just a temporary glitch. We’ve seen such glitches a couple of times before. I don’t really know what could have changed that early in the installation process.

Issue #203, opened 2022-11-18
[ocaml5-issue] Windows failures on threadomain


jmid:
We have also observed failures on Windows on the `threadomain` test combining `Domain`s and `Thread`s.

Here is a fresh occurrence on `5.0.0~beta1`:
https://github.com/jmid/multicoretests/actions/runs/3487203320/jobs/5834552533
```
random seed: 152027115
generated error fail pass / total     time test name

[ ]    0    0    0    0 /  500     0.0s Mash up of threads and domains
File "src/threadomain/dune", line 15, characters 0-113:
15 | (rule
16 |  (alias runtest)
17 |  (package multicoretests)
18 |  (deps threadomain.exe)
19 |  (action (run ./%{deps} --verbose)))

(cd _build/default/src/threadomain && ./threadomain.exe --verbose)
Command exited with code -1073741819.
[ ]    0    0    0    0 /  500     0.0s Mash up of threads and domains (generating)
random seed: 450754698
```

This exits with a weird exit code `-1073741819`

jmid:
This was previously observed on #179 - also on the Windows `5.0.0~beta1` combination and with the same symptom of a weird exit code:
https://github.com/jmid/multicoretests/pull/179#issuecomment-1315796191

shym:
Following @dra27, I wrote a dummy executable to trigger a segfault:
```
PS> .\a
PS> echo $LASTEXITCODE
-1073741819
```
(-1073741819 is 0xc0000005, by the way).

dra27:
> (-1073741819 is 0xc0000005, by the way).

Which is access violation / segfault 👍

jmid:
Saw what looks like a deadlock and a 6 hour timeout on `threadomain` on Windows `5.0.0+trunk` - with none of the previous tests haven taken substantial amounts of time:
https://github.com/jmid/multicoretests/actions/runs/3497938069/jobs/5857667163
```
random seed: 528980161
generated error fail pass / total     time test name

[ ]    0    0    0    0 /  500     0.0s Mash up of threads and domains
[ ]    0    0    0    0 /  500     0.0s Mash up of threads and domains (generating)
[ ]   52    0    0   52 /  500    60.4s Mash up of threads and domains
[ ]  106    0    0  106 /  500   [120](https://github.com/jmid/multicoretests/actions/runs/3497938069/jobs/5857667163#step:8:121).6s Mash up of threads and domainsTerminate batch job (Y/N)? 
[...]

PR #285, opened 2023-01-20
Streamline CI

shym:
As a follow-up to the previous PR, this one adds two new CI workflows for the bytecode version on Windows and the corresponding badges. It also removes some remaining mentions of `rc1`, drops the alpha opam repository (which is no longer necessary with the release of OCaml 5) and logs the opam repositories from which dependencies are pulled (to make sure the multiple opam repositories are indeed providing exactly the packages they are expected to provide, and not conflicting with the main default one).


jmid:
PR looks great - thanks!
I'll let the CI run finish and then merge. The early macOS failure looks like a temporary network hiccup.

jmid:
This looks a lot greener! :smiley:  ... but it does include a Win bytecode crash! :grimacing: 
If nothing else, this looks relatively easy to trigger consistently? :thinking:

shym:
Indeed, I see many many deadlocks, maybe the bytecode version is the right one to try and finally corner that bug (if it’s indeed #203 since PR11846 seems to fix-or-hide it in native)

jmid:
OK, so we have
- a windows bytecode crash
- a windows threadomain timeout

We cannot blame the PR for any of these though.
Do you agree to merge and then investigate these in parallel @shym?

shym:
Indeed, I think we can merge it! :+1:

PR #288, opened 2023-01-25
Separate and label internal tests

jmid:
This PR
- separates the internal `qcheck-lin`, `qcheck-stm`, and `qcheck-multicoretests-util` tests into a separate `test/` directory and
- labels them with their appropriate package name

This means
- we can adjust the CI workflow to run them first in a separate step
- they can be included in released packages and run during installation with `--with-test`

For now I prefer to not include the `ref` and `conc_list` tests in the above category, since we are occasionally seeing failures in these (over `Thread` at least) and I would hate for a user to have to understand why a flaky test made their package installation `--with-test` fail... :grimacing:


shym:
It’s a nice clarification of the roles of the various tests, indeed :+1:

jmid:
CI is green modulo a bunch of Windows `threaddomain` failures (1 crash and 4 timeouts) which are all unrelated, so I'll merge.

PR #289, opened 2023-01-25
Add a (negative) Lin test for Dynlink module

shym:
This adds a (preliminary) Lin test for the `Dynlink` module, only testing three functions at the moment.
Its first run triggered a segfault on Windows in CI.


jmid:
This is good to go! The one CI failure is unrelated: threadomain on Windows, so I'll merge :smiley:

PR #291, opened 2023-01-25
Rewrite the CI workflows to run the test suite on a specific compiler (specific commit or branch)

### LINKED ISSUE #286: CI focused testing based on compiler commit
I just tried the CI options to test a compiler based on a specific commit `'7942816c0'`
This failed because after downloading the directory is named with the full hash: `ocaml- 7942816c058b0ad2ca168e9f4063eaaa8240ba79/` - see e.g. https://github.com/ocaml-multicore/multicoretests/actions/runs/3968857974/jobs/6802577780

I then switched to specifying a full git hash in c6c16d6 - which fails since it didn't pin the 5.1-version of `ppxlib`
https://github.com/ocaml-multicore/multicoretests/actions/runs/3968925696/jobs/6802734116

None of this was unsolvable - but I'm likely to forget again and hence writing it here to remind my future self :sweat_smile: 

For the future, we could
- document in the `common.yml` workflow file that full hashes are expected
- `pin` the 5.1-version of `ppxlib` for the commit-based testing as we are probably most likely to do so on 5.1 commits (and the 5.1 ppxlib version is backwards compatible)

@shym What do you think? :thinking:

shym:
Instead of installing a standard compiler package and later updating that opam switch with a new compiler version, generate an opam package configuration into a `local` OPAM repository
This allows to directly install the intended compiler version (a specific compiler commit, for instance) and provides caching of that compiler by setup-ocaml

The last commit of that PR should not be merged into `main` but it allows simply to test the (moving) compiler 5.0.1. (There is an OPAM package that would allow to test that compiler simply on macOS and Linux but not on Windows.)

This hopefully closes #286.

### DIFF
--- a/.github/opam/custom/repo
+++ b/.github/opam/custom/repo
@@ -1,0 +1,1 @@
+opam-version: "2.0"
--- a/.github/opam/ocaml-variants.opam.unix
+++ b/.github/opam/ocaml-variants.opam.unix
@@ -1,0 +1,108 @@
+opam-version: "2.0"
+name: "ocaml-variants"
+synopsis: "Custom compiler __OCAML_COMPILER_FULL_VERSION__"
+maintainer: "platform@lists.ocaml.org"
+authors: [
+  "Xavier Leroy"
+  "Damien Doligez"
+  "Alain Frisch"
+  "Jacques Garrigue"
+  "Didier Rémy"
+  "Jérôme Vouillon"
+]
+license: "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+homepage: "https://ocaml.org"
+bug-reports: "https://github.com/ocaml/opam-repository/issues"
+depends: [
+  "ocaml" {= "__OCAML_OPAM_PACKAGE_VERSION__" & post}
+  "base-unix" {post}
+  "base-bigarray" {post}
+  "base-threads" {post}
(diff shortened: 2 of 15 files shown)


shym:
This also opens up the possibility to add workflows to test more regularly 5.0.1. I’d suggest using a cron trigger (once a week, maybe?) rather than on every push and PR. Any opinion about this, @jmid?

jmid:
OK, I've now had a look at this. Having opam file templates that can be filled out to enable faster testing + caching of custom compilers makes good sense to me! :smiley: :+1: 

I suspect there is some risk associated with having to maintain the two templates and keeping them in sync with upstream opam files? I suppose a mismatch could go either way (fails only on copy - or fails only on custom repo).
I guess it is a trade-off: speed vs. maintenance. What do you think?

shym:
I just updated this PR with 3 commits replacing the “try it on 5.0.1” one.
The last new commit adds the idea of triggering workflows on 5.0.1 at every beginning of the week, as a way to test it regularly (but not as constantly as we do on 5.0.0 and trunk) but also to check that the new “install a custom compiler” keeps working. I think that partly addresses the risk of not keeping up with upstream opam files.
(By the way, `ocaml-variants.opam` in `ocaml/ocaml` does seem to move only slowly, so I feel this is rather safe)

PR #295, opened 2023-02-02
Run cleanup in the presence of exceptions

jmid:
While looking at the many similar `Lin` and `STM` properties it struck me that `cleanup` wouldn't always be run if an exception would be raised underway. This PR attempts to rectify the situation by using our old friend, the `result` type :slightly_smiling_face: 

The affected properties/libraries are:
- `Lin_thread`
- `Lin_effect`
- `STM_domain`
- `STM_sequential`

For the latter two, I dusted off and included an old test case from the reverted #143 - where I was trying to do the same, but messing up in the exception handling :sweat_smile: :shrug:


jmid:
CI summary: 2 threadomain Windows crashes and 1 Windows Dynlink test failure

jmid:
Rebased on `main` to resolve merge conflicts and `CHANGES` entry added

PR #296, opened 2023-02-07
Skip Dynlink tests under Windows

shym:
The current implementation of Dynlink under Windows is not thread-safe so skip it until it is fixed


jmid:
PR looks good to me! :+1:

shym:
As the test has not yet been really exercised against Cygwin, where the dynlink code is substantially different, I updated the test to keep it enabled there.

PR #297, opened 2023-02-09
Rename the bounds of the tree generators

shym:
After checking in the _Introduction to Algorithms_, use the terms height and degree to bound the size of the generated trees and modify the code to match the definitions of those terms

### DIFF
--- a/src/domain/domain_spawntree.ml
+++ b/src/domain/domain_spawntree.ml
@@ -27,17 +27,17 @@ type cmd =
   (*| Join*)
   | Spawn of cmd list [@@deriving show { with_path = false }]
 
-let gen max_depth max_width =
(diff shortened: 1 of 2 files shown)


jmid:
LGTM! :+1: 
(I'll add this to the stack of mergable PRs...)

jmid:
Actually, I just remembered that I shamelessly copy-pasted and adapted the approach to stress test 5.0 `Thread`s too...
Can I ask you to include the same renaming for `src/thread/thread_createtree.ml`? :pray:

shym:
The commit modifies both `**/*tree.ml` files similarly. Is there another test I should have modified?

jmid:
Sorry, while scrolling I missed that you already updated both of them! :facepalm:

PR #301, opened 2023-02-27
Add opam installation instructions

jmid:
This PR
- clarifies `opam` installation steps to reflect that we've now released the two libraries and
- tones down the 'experimental' signals in the README text


README.md
```diff
@@ -40,7 +41,14 @@ opam update
 opam switch create 5.0.0
 ```
 
-Using `opam` you can now `pin` and install them as follows:
+Using `opam` you can install the two testing libraries by running one
```
shym, line 44: I’m just wondering: could this formulation suggest you should install only of them?
I’d suggest the following phrasing: “The two testing libraries are available as OPAM packages `qcheck-lin` and `qcheck-stm` in the standard repository, so they can be installed in the usual way:”

shym (approved):
:+1:

README.md
jmid: Thanks - rephrased in e6d18a6

jmid:
@shym : I see the two 32-bit `trunk` builds failing with the same error:
```
/usr/bin/ld: skipping incompatible /usr/lib/x86_64-linux-gnu/libzstd.so when searching for -lzstd
```
I suspect this is caused by https://github.com/ocaml/ocaml/pull/12006 which just got merged that adds a Zstd dependency - where we then pick up the 64-bit Zstd version.

jmid:
Merging as the README changes are unrelated to the 32-bit CI breaking.

PR #302, opened 2023-02-28
CI: Install 32-bit libzstd for 32-bit trunk runs

shym:
Trunk OCaml now uses libzstd to compress during marshalling so we explicitly install the 32-bit version of the library in 32-bit runs and modify LDFLAGS to ensure that library is found during linking

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -112,6 +112,12 @@ jobs:
         id: presetup
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


jmid:
CI summary:
- on macOS `trunk` the `Lin` `Out_channel` test failed to trigger the expected issue
- on Windows bytecode `trunk` threadomain failed to terminate in time

Both are unrelated, so I'll merge - thanks! :pray:

Issue #306, opened 2023-03-09
[ocaml5-issue] Segfault in Buffer test (5.0, 32bit)


shym:
Observed in a run on released OCaml 5.0 in 32 bits, a SEGV in the buffer test:
https://github.com/shym/multicoretests/actions/runs/4367430742/jobs/7638729780#step:20:1204

```
random seed: 470817338
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s STM Buffer test sequential
[ ]    0    0    0    0 / 1000     0.0s STM Buffer test sequential (generating)
[✓] 1000    0    0 1000 / 1000     1.9s STM Buffer test sequential

File "src/buffer/dune", line 4, characters 7-16:
4 |  (name stm_tests)
           ^^^^^^^^^
[ ]    0    0    0    0 / 1000     0.0s STM Buffer test parallel
(cd _build/default/src/buffer && ./stm_tests.exe --verbose)
Command got signal SEGV.
```

jmid:
Cool! :sunglasses: :+1: 
I suggest hacking the tested property to print the triple as a side-effect.
That way, you can get a better sense of the input causing it - and perhaps write an STM-independent reproducable.

The old `diff` from my last hack follows below.
(we should consider incorporating some way to obtain something like it that doesn't involve hacking on the library...)
```
diff --git a/lib/STM_domain.ml b/lib/STM_domain.ml
index b8bd1c7..fc11554 100644
--- a/lib/STM_domain.ml
+++ b/lib/STM_domain.ml
@@ -17,8 +17,11 @@ module Make (Spec: Spec) = struct
     let res_arr = Array.map (fun c -> Domain.cpu_relax(); Spec.run c sut) cs_arr in
     List.combine cs (Array.to_list res_arr)
 
-  let agree_prop_par (seq_pref,cmds1,cmds2) =
+  let ch = open_out "input_log.txt"
+
+  let agree_prop_par ((seq_pref,cmds1,cmds2) as trip) =
     assume (all_interleavings_ok seq_pref cmds1 cmds2 Spec.init_state);
+    Printf.fprintf ch "Triple: %s\n%!" Print.(triple (list Spec.show_cmd) (list Spec.show_cmd) (list Spec.show_cmd) trip);
     let sut = Spec.init_sut () in
     let pref_obs = interp_sut_res sut seq_pref in
     let wait = Atomic.make true in
[...]

PR #309, opened 2023-03-14
Add the Buffer.add_string race condition to the list of issues

shym:
Add the resolution of the bug tracked in #306

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,14 @@ property can be done in two different ways:
 Issues
 ======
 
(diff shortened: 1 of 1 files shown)


jmid:
Quite on point, the CI triggered a `Buffer` segfault on 32-bit 5.0 :grin: 
There was also a Win 5.0 Threadomain timeout and a flaky GitHub actions installer failure.
Merging.

PR #311, opened 2023-03-20
Add the flexdll race condition to the list of issues

shym:
Add the race condition in accesses to the global variables storing the last error in flexdll to the list of issues found (not yet fixed)


jmid:
Thanks!

PR #312, opened 2023-03-21
Escape bytes when printing in STM

shym:
Same problem as in #281 for `bytes` (see for instance this [run](https://github.com/shym/multicoretests/actions/runs/4469161241/jobs/7850880362#step:21:947)), so same solution.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - ensure `cleanup` is run in the presence of exceptions in
   - `STM_sequential.agree_prop` and `STM_domain.agree_prop_par`
   - `Lin_thread.lin_prop` and `Lin_effect.lin_prop`
+- #312: Escape and quote `bytes` printed with `STM`'s `bytes` combinator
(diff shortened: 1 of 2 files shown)


jmid (approved):

jmid:
Come to think of it: Can you add a `Changes` entry for this, similar to the one for `string`?
It would be nice to include in the release notes for the next version :pray:

shym:
Done.

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 - ensure `cleanup` is run in the presence of exceptions in
   - `STM_sequential.agree_prop` and `STM_domain.agree_prop_par`
   - `Lin_thread.lin_prop` and `Lin_effect.lin_prop`
+- #312: Escape and quote strings printed with `STM`'s `bytes` combinator
```
jmid, line 8: `Escape and quote bytes`? :grin:

PR #313, opened 2023-03-22
Split the CI for Cygwin into 2 jobs rather than 2 workflows

shym:
I finally found out that stating that the second job `needs` the first, it is not skipped. This allows us to get one workflow for the whole test suite on Cygwin (but unfortunately it can take quite some time, with 2 × 6 hours of timeout).


jmid (approved):
Very nice - and nice code clean-up! :smiley: 
(per tradition I'll let the CI finish before merging)

jmid:
All green - for the first time in a while! I'll merge immediately :smile:

PR #314, opened 2023-03-23
Disable the negative Out_channel tests on Windows and macOS

shym:
These tests are not triggering consistently on those platforms (and often take a lot of time) so this PR proposes to disable them at least for now, to reduce the noise of CI failures.
Maybe the frequencies could be tweaked so that they trigger most consistently? :thinking:

### DIFF
--- a/src/io/dune
+++ b/src/io/dune
@@ -23,7 +23,9 @@
  (package multicoretests)
  ;(flags (:standard -w -27))
  (libraries qcheck-lin.domain lin_tests_dsl_common_io)
- (action (run %{test} --verbose))
+ (action
+  (setenv OCAML_SYSTEM %{system}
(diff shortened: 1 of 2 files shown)


jmid:
> These tests are not triggering consistently on those platforms (and often take a lot of time) so this PR proposes to disable them at least for now, to reduce the noise of CI failures. 

Reducing the noise would indeed be very welcome!

> Maybe the frequencies could be tweaked so that they trigger most consistently? :thinking:

I was wondering about that myself! Is there a particular shape triggering these failures on macOS and Windows, e.g., if one compares the failures (once they fail)? If we could have them trigger more consistently, that would also prevent them from completing and hence speed them up... :thinking:

jmid:
Closing this in favor of our #319 and #326 combination.

PR #316, opened 2023-03-23
Fix rep_count in STM_thread

jmid:
I accidentally discovered that the two `rep_count`s in `STM_thread` were different:
- `agree_test_conc` would perform 100 repetitions whereas
- `neg_agree_test_conc` would perform only 25 repetitions

I think this is simply a copy-paste error.

I wonder whether this would allow us to trigger `STM_thread` test errors more consistently... :thinking:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Next version
 
+- #316: Fix `rep_count` in `STM_thread` so that negative and positive
+  tests repeat equally many times
 - #318: avoid repetitive interleaving searches in `STM_domain` and `STM_thread`
 - #312: Escape and quote `bytes` printed with `STM`'s `bytes` combinator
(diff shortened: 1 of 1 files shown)


jmid:
Rebased on `main` and added a CHANGES entry

jmid:
CI summary:
- The `Out_channel` test failed to trigger on Windows bytecode and macOS `trunk`
  - it also took 9440.3s on macOS causing a time out and appears to cause a timeout on macOS trunk
- Cygwin `trunk` part 1 timed out with a number of different tests taking a bit long
- `Lin DSL Weak HashSet test` took an extraordinary 14177.8s to shrink on Windows bytecode 5.0.0 and 9009.0s on Cygwin `trunk` part 2 both causing a timeouts
- `threadomain` dead locked on Windows trunk

None of these are related to this PR so I'll merge.

PR #318, opened 2023-03-24
Avoid repetitive interleaving search in STM_domain

jmid:
In STM_domain we use `assume (all_interleavings_ok seq_pref cmds1 cmds2 Spec.init_state)` to ensure that all `cmd` preconditions in a triple are satisfied under all interleavings.
I realize we are checking the precondition repeatedly - once for each `Util.repeat` iteration! :scream: 
By default, that means we are exploring that same exponential space 25 times whereas 1 one would do.
The PR proposes to move the precondition check outside `agree_prop_par` and the `Util.repeat` loop for both `STM_domain` and `STM_thread`.

Overall I was expecting grand CI speed-ups from this - and I encountered a few of those, e.g., `Linux trunk` in 22m31s!
Many runs however didn't speed up because of long running IO tests... I'm hoping that this together with frequency adjusted or disabled IO tests (#314) will collectively provide a solid boost in CI testing time :crossed_fingers:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Next version
 
+- #318: avoid repetitive interleaving searches in `STM_domain` and `STM_thread`
 - ensure `cleanup` is run in the presence of exceptions in
   - `STM_sequential.agree_prop` and `STM_domain.agree_prop_par`
   - `Lin_thread.lin_prop` and `Lin_effect.lin_prop`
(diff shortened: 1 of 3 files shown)


jmid:
Rebased on `main`

jmid:
CI summary:
- one Windows bytecode `trunk` crash on threadomain
- a macOS `trunk` `Out_channel` test issue failing to trigger
- a macOS `trunk` `Out_channel` test causing a timeout

None of these are related to the current PR changes, so I'll merge.

PR #319, opened 2023-03-24
Out_channel frequency adjustment

jmid:
I was looking at the `Out_channel` tests and noticed something: most of the tested operations are returning `unit` (in contrast to `In_channel` where most of them return observable values).
This decreases the chance of observing sequential consistency failures about the tested `Out_channel.t`.

This PR therefore increases the frequency of non-unit-returning functions, `pos` and `length` in `Out_channel`.
It works reasonably well: focused tests repeated 10 times triggers the expected failures on Windows and macOS as expected within the first 100-200 generated test input. We should consider this as an alternative to #314.

(Note: CI is failing on `trunk` while compiling `ppxlib` due to an AST-changing PR earlier today. The Windows `trunk` version is however based on another branch which hasn't been updated yet AFAICS)

### DIFF
--- a/src/io/lin_tests_dsl_common.ml
+++ b/src/io/lin_tests_dsl_common.ml
@@ -67,8 +67,8 @@ module OCConf : Lin.Spec = struct
     (* val_ "Out_channel.with_open_gen"    Out_channel.with_open_gen    (open_flag list @-> int @-> string @-> (t @-> 'a) @-> returning 'a) ; *)
 
     val_ "Out_channel.seek"             (lift Out_channel.seek)             (t @-> int64 @-> returning_or_exc unit) ;
-    val_freq 2 "Out_channel.pos"        (lift Out_channel.pos)              (t @-> returning_or_exc int64) ;
-    val_freq 2 "Out_channel.length"     (lift Out_channel.length)           (t @-> returning_or_exc int64) ;
+    val_freq 3 "Out_channel.pos"        (lift Out_channel.pos)              (t @-> returning_or_exc int64) ;
+    val_freq 3 "Out_channel.length"     (lift Out_channel.length)           (t @-> returning_or_exc int64) ;
     val_ "Out_channel.close"            (lift Out_channel.close)            (t @-> returning_or_exc unit) ;
     val_ "Out_channel.close_noerr"      (lift Out_channel.close_noerr)      (t @-> returning unit) ;
     val_ "Out_channel.flush"            (lift Out_channel.flush)            (t @-> returning_or_exc unit) ;
--- a/src/io/lin_tests_dsl_common.ml
+++ b/src/io/lin_tests_dsl_common.ml
@@ -67,8 +67,8 @@ module OCConf : Lin.Spec = struct
     (* val_ "Out_channel.with_open_gen"    Out_channel.with_open_gen    (open_flag list @-> int @-> string @-> (t @-> 'a) @-> returning 'a) ; *)
 
     val_ "Out_channel.seek"             (lift Out_channel.seek)             (t @-> int64 @-> returning_or_exc unit) ;
-    val_ "Out_channel.pos"              (lift Out_channel.pos)              (t @-> returning_or_exc int64) ;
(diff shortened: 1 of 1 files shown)


jmid:
CI is all green (modulo the failing `ppxlib` errors) except for the Cygwin runs:
- Here the repeating happens on each of the two halves (OK I guess)
- The first one fails though, due to excessive shrinking: https://github.com/ocaml-multicore/multicoretests/actions/runs/4515900237

In conclusion, the frequency adjustment to more consistently trigger an `Out_channel` error seems to work. :tada: 
However, a new problem of occasional excessive shrinking is surfacing (which we are also observing on Weak Hashsets).

shym:
I had also tried to see if adjusting the frequencies could help in [this commit](https://github.com/shym/multicoretests/commit/29c7236689f5dd5b3f202320fe1b212049ad6152). Instead of using some objective criteria (not returning `unit`), I tried to favour the functions naturally conflicting. The results don’t seem so different from what you got, I think. Interesting.

shym:
> (Note: … The Windows trunk version is however based on another branch which hasn't been updated yet AFAICS)

On Windows and Cygwin, something weird happens: building `ppxlib` fails, (as we should expect) but the test is still run?!? (the only things I can imagine would be that the VM is not properly cleaned up between runs...) As it shouldn’t build in that setting, I have no idea what is actually running :thinking: I would at least say we shouldn’t draw conclusions on what happens in those windows builds.

jmid:
> On Windows and Cygwin, something weird happens: building ppxlib fails, (as we should expect) but the test is still run?!?

Oh, I mistakenly thought they were based on a patched Windows/Cygwin branch (which would explain why they weren't broken by `trunk`'s AST change.

The `Lin` test it is focusing on doesn't require `ppxlib`, so in principle it should still be runnable despite a broken `ppxlib` install...

jmid:
Rebased on top of a freshly merged #320...

jmid:
CI error summary:
- one macOS install failure
- one Windows `trunk` test not triggering
- several timeouts on Cygwin/Windows/macOS due to excessive shrinking

jmid:
Rebased on `main` and removed the 'focus commit' to see how this influences CI in general.

jmid:
CI results are a mixed bag, mostly unrelated to this PR:
- `src/neg_tests/lin_tests_thread_conclist.exe` triggered an error under Windows bytecode
- `src/neg_tests/stm_tests_thread_ref.exe` failed to trigger on macOS 5.0
- the `Out_channel` test failed to trigger on macOS trunk twice (for push + pull request)
- Cygwin trunk part 1 timed out due to several tests taking longer
- `threadomain` dead locked on Windows bytecode trunk twice (for push + pull request) and on Windowd 5.0
- `Lin DSL Weak HashSet` test takes a whopping 16423.5s on Cygwin trunk part2 causing a timeout
- `Lin ref int64 test with Thread` takes 9399.1s also on Cygwin trunk part2 causing a timeout

The second item of the above should be taken care of by #316 
I'm surprised to still see two failures to trigger the `Out_channel` issue on macOS though...

jmid:
With #316 merged I wanted to give this another spin.

I've thus
- bumped the frequency count of `pos` and `length` from 2 to 3 in an attempt to trigger more consistently under macOS
- merged `main` into the branch to reduce noise from other test failures

Now: :popcorn:

jmid:
Latest CI summary:
- a threadomain crash on Windows `trunk` and a deadlock on Windows bytecode `trunk`
- a failure to trigger the `Out_channel` issue on macOS `trunk`
- a timeout on Cygwin `trunk` part2 partly due to `Lin ref int64 test with Thread` taking 8041.6s

jmid:
@shym it seems the frequency adjustment lets us trigger the issue more consistently on Windows.
macOS is however still not consistent! :grimacing: 
A suggestion: we combine this PR and #314. That way we only need to disable the test on macOS. WDYT?

shym:
Sounds good! #326 is there to disable the mac test on this PR.

jmid:
Great, I'll merge this then :+1:

jmid:
Doh! I was too fast and missed that you wanted to merge #326 into this one :facepalm:

PR #320, opened 2023-03-27
Fixes small issues in CI

shym:
Runs with `ppxlib` no longer building with `trunk` revealed two issues that were unnoticed in the Windows workflows


jmid:
CI is green modulo a Windows `Out_channel` test that failed to trigger under Windows trunk.
I'm eager to merge to try out a rebased #319...  :smiley:

PR #323, opened 2023-04-06
Add Lin tests for Base.Buffer

shym:
Add Lin tests for Base.Buffer. This is split into two commits to keep track of the functions that can raise exceptions when used in parallel (while they shouldn’t in a sequential use).

Segfaults happened on 5.0.0 in CI, for instance [there](https://github.com/shym/multicoretests/actions/runs/4619228913/jobs/8167695430#step:20:799):
```
random seed: 255066454

[ ]    0    0    0    0 / 1000     0.0s Lin DSL Base Buffer test with Domain
/home/runner/work/_temp/cfbbcd46-3b2b-4f6b-b063-35a9068fc155.sh: line 2: 3690717 Segmentation fault      (core dumped) opam exec -- dune exec "$ONLY_TEST" -- -v
[ ]    0    0    0    0 / 1000     0.0s Lin DSL Base Buffer test with Domain (generating)
```
This is hopefully another symptom of ocaml/ocaml#12103 which was fixed on `trunk` by ocaml/ocaml#12104.

### DIFF
--- a/src/base/buffer/BaseOrNoBase.fallback.ml
+++ b/src/base/buffer/BaseOrNoBase.fallback.ml
@@ -1,0 +1,3 @@
+module Buffer = Stdlib.Buffer
+
+let _ = failwith "This is using a placeholder for compilation, do not run!"
--- a/src/base/buffer/BaseOrNoBase.passthrough.ml
+++ b/src/base/buffer/BaseOrNoBase.passthrough.ml
@@ -1,0 +1,1 @@
+module Buffer = Base.Buffer
--- a/src/base/buffer/dune
+++ b/src/base/buffer/dune
@@ -1,0 +1,12 @@
+(executable
+ (name lin_tests_dsl)
+ (libraries qcheck-lin.domain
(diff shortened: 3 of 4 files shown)


shym:
For some reason, `enabled_if` seemed to work locally but is not enough to tell `dune` not to compile the test in CI, so I updated the PR with a workaround using a `select ... from`. It didn’t seem to work in a `test` stanza, though, so I unfolded it into an `executable` and a `rule`.

shym:
(Runs on `trunk` will break, it was updated to 5.2, `ppxlib` must follow suite)

jmid:
I've given these tests a spin on 5.1 and trunk with this branch and cannot reproduce with a `repeats` count of 50: https://github.com/ocaml-multicore/multicoretests/tree/shym-base-buffer
This indicates that the issue is indeed fixed by ocaml/ocaml#12104.

jmid:
Closing this for now, as we abandoned the idea of testing `Base`.
Thanks for the PR and for confirming that this could be useful too! :smiley:

PR #328, opened 2023-04-14
Update issues

jmid:
This PR updates the list of issues
- marking the `Weak` `Hash` `set` crash as solved
- adding the MingW `Sys.rename` issue


jmid:
CI summary:
- two threadomain crashes on Windows bytecode trunk and 5.0.0
- a threadomain deadlock on Windows bytecode trunk
- Linux 5.0.0 timing out since `Lin DSL Weak HashSet test` took a whopping 8505.5s

PR #331, opened 2023-04-19
Add Util functions to make tests with logging of generated test values

shym:
This has been useful quite a few times for us, for instance to investigate deadlocks or livelocks, and build standalone reproducers. This might be more generally useful.
This is still a draft, with the following points to fix:
- documentation for the new functions
- possibly rename those functions to explicitly mention `triple`
- possibly add slightly more generic versions, not tied to triples, so that all our `Test.make{,_neg}` can be replaced by these wrappers
- maybe share the code between the `make` functions.

### DIFF
--- a/lib/STM_domain.ml
+++ b/lib/STM_domain.ml
@@ -66,7 +66,7 @@ module Make (Spec: Spec) = struct
     let rep_count = 25 in
     let seq_len,par_len = 20,12 in
     let max_gen = 3*count in (* precond filtering may require extra generation: max. 3*count though *)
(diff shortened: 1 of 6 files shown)


jmid:
Cool! I was thinking about this yesterday too :+1: 

Logs is a nice and popular library for this kind of logging: https://erratique.ch/software/logs (sans dependencies).
AFAIU, it works through callbacks to a 'reporter'. By default, a "nop-reporter" does nothing and hence is relatively cheap.
To have things printed, one then just need to set another reporter.
I was wondering whether it would be handy to use for us? :thinking:

PR #333, opened 2023-04-20
Series of misc improvements to CI configuration

shym:
Small improvements to CI that have been pending but not deserving PRs on their own.
The simplifications to 32-bit-run configurations are follow-ups from tests for ocaml/ocaml#12190.


jmid (approved):
Great - LGTM - thanks! :smiley: 
 (I'll let the CI finish before merging)

jmid:
Hurrah! There was a Windows bytecode trunk timeout, saving us 2 hours of CI, so this is already showing its value... :tada: 
There are still two Cygwin workflows underway. It would be nice if we could pinpoint the reason for the long runtimes of these... :thinking:

PR #334, opened 2023-04-21
CI: Catch apt-get --yes install just as apt-get install

shym:
Following ocaml/setup-ocaml#658, the invocation of apt-get changed, which broke linux CI


jmid (approved):
Wonderful, thanks a bunch! :pray:

jmid:
In the interest of moving forward, I'll go out on a limb and merge now - rather than wait 10h for the Cygwin runs... :sweat_smile:

PR #336, opened 2023-04-21
CI: Disable sudo/brew overriders

shym:
Drop the list of workflow specific packages to install Move the installation of *:i386 packages to the common workflow
This is trading a little bit of CI time and downloads for better robustness to setup-ocaml updates


jmid:
Only the dog-slow Cygwin runs remain while the rest are green - so I'll go ahead and merge.

PR #337, opened 2023-04-25
Add missing bytes combinators to Lin

### LINKED ISSUE #83: Missing `bytes` combinator in lin_api
lin_api.ml is missing a `bytes` combinator.

It would make sense to fix c-cube/qcheck#83 first - and then build upon that

jmid:
This PR fixes #83 - an old omission that I had forgotten about until #323 reminded me.

Because of value restriction, code with a call cannot have polymorphic type. This is the reason I've let-bound `QCheck.bytes_small_of QCheck.Gen.printable` above.

As a small bonus, I include a couple of minor documentation typos I spotted while copy-pasting from the `string` combinators... :sweat_smile:


jmid:
Corrected a small nit in 3d9cf64 and rebased on main.

jmid:
CI summary:
- a Windows bytecode 5.1 threadomain crash
- a Cygwin trunk part1 timeout due to long-running Bigarray/Bytes/Domain.spawn-join/Atomic/Hashtbl tests
- two Windows trunk bytecode threadomain timeouts
- a WIndows trunk threadomain timeout
- a Cygwin trunk part2 timeout due to excessive `Lin DSL Weak HashSet test with Domain` shrinking
- a Cygwin trunk part2 timeout due to excessive `Lin ref int64 test with Thread` shrinking

None of these have to do with the PR, so I'll merge

PR #342, opened 2023-05-11
Remove ppx generators

### LINKED ISSUE #341: Eliminate the ppxlib dependency
ppxlib-derivers are nice as they save us some boilerplate code for generators, `show` functions, and equality predicates.
On the other hand, they introduce a hard dependency which makes the test suite fragile to AST changes, causing CI etc. to break until `ppxlib` has been patched. It would thus make sense to bite the bullet and hand-write this boilerplate and thereby eliminate the dependency.

shym:
This PR removes the ppx derivers for `show`, `eq` and `qcheck`.

Open questions:
- should we remove also derivers in `doc/`?
- should we use `(=)` instead of those `equal_` functions?
- we could improve the pretty-printers to get closer to standard OCaml printers and/or to the derived ones (opening boxes, etc.) The code is structured so that this should be easy to do.

Closes #341.


jmid (approved):
One word: Awesome! :smile: :tada: 

> should we remove also derivers in doc/?

I actually thought the examples were run in the CI - but I realize now that they are not.
Based on the above, I was thinking of using `ppx_deriving` in the text (to keep it short) but including a commented `show` etc in the code
```
 (* Note: you can omit the following definition if you uncomment 
      (preprocess (pps ppx_deriving.show))
    in the dune file *)`
```
Since we don't run the examples, I think it is fine to leave it as is though.

> should we use (=) instead of those equal_ functions?

Polymorphic equality would certainly make things a bit shorter, but there are also arguments against using it entirely. I think what you have done is very reasonable :+1: 

> we could improve the pretty-printers to get closer to standard OCaml printers and/or to the derived ones (opening boxes, etc.) The code is structured so that this should be easy to do.

That would be nice to have(TM) but not strictly needed ATM.

I also realize our expect outputs were already a bit inconsistent in we would print
`(Remove -605797133) : Some (-605797133)` which is now 
`Remove -605797133 : Some (-605797133)` - where we have a constructor with a negative `int` argument on either side of the colon.

This is just me getting distracted and doesn't mean anything for the PR.
A CHANGES entry to announce the new `Util` sub-modules `Pp` and `Equal` would perhaps be in order...

shym:
> I actually thought the examples were run in the CI - but I realize now that they are not.

I thought so too, I expected them to be run at least in the documentation workflow, but I don’t think they are. Anyway, keeping the ppxlib dependency for examples that are run on current stable didn’t feel such a big deal.

> Since we don't run the examples, I think it is fine to leave it as is though.

:+1:

> I also realize our expect outputs were already a bit inconsistent in we would print `(Remove -605797133) : Some (-605797133)` which is now `Remove -605797133 : Some (-605797133)` - where we have a constructor with a negative `int` argument on either side of the colon.

Well-spotted! The pretty-printers for numbers should be fixed now, and use parentheses when required for negative values.

> A CHANGES entry to announce the new `Util` sub-modules `Pp` and `Equal` would perhaps be in order...

I added that.

Along those fixes, I also dropped the use of optional argument, as it was more cumbersome than I had hoped. The few cases where it allowed the implicit uses of the optional value were making it annoying when you define functions that must have that optional argument but they will ignore it (it is the case of pretty-printers for sum types when no constructor has an argument, for instance).

I also added documentation to the `Util.Pp` module, since it is exposed and is useful to users of the library.

jmid:
Still LGTM :smiley::+1: 
:heavy_plus_sign: for removing the optional parameters.
It is also my experience from QCheck, that they don't always work well together with combinators.

jmid:
CI summary:
- Windows 5.1 failed to trigger the Out_channel issue
- Windows bytecode trunk triggered the threadomain lock issue #203 
- Cygwin trunk, part2 timeout, due to excessive `Lin DSL Weak HashSet test with Domain` shrinking

None of these are related to the PR, so I'll merge :+1:

Issue #344, opened 2023-05-15
32-bit pre-setup fails


jmid:
We're just starting to see the 32-bit pre-setup fail, e.g.:
https://github.com/ocaml-multicore/multicoretests/actions/runs/4977631855/jobs/8906911507
```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 libc6:i386 : Depends: libgcc-s1:i386 but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
Error: Process completed with exit code 100.
```

I wonder if some Ubuntu package had it's dependencies changed over the last few day :thinking:

PR #346, opened 2023-05-15
Opam installation updates

### LINKED ISSUE #345: opam workflow fails on 5.1
The weekly workflow to check opam installation of the 3 testing packages fails on 5.1+trunk
https://github.com/ocaml-multicore/multicoretests/actions/runs/4975788465/jobs/8903271370
due to an insufficient ppxlib version (we don't pin a particular trunk or 5.1-patched ppxlib in that workflow).

I suspect this will solve itself with the merge of #342 - but it might be nice to check.

jmid:
This fixes #345 

When checking to see whether the ppx avoiding in #342 would fix the failing `opam install` from #345, I discovered that the #342's refactoring actually broke the 4.13 support just added in #340...

This PR thus 
- writes out the `Seq.equal` definition contributed in #340 
- updates the `opam install` workflow to run on all changes

The install workflows are quite fast, so I favor protecting against such breakage again now that we actually support 4.13 and 4.14.

### DIFF
--- a/.github/workflows/opam.yml
+++ b/.github/workflows/opam.yml
@@ -4,11 +4,7 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
-on:
-  schedule:
-    # Every Monday morning, at 1:11 UTC
-    - cron: '11 1 * * 1'
-  workflow_dispatch:
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build-and-test:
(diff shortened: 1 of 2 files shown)


jmid:
CI summary:
- The 32 bits test are failing as noted in #344 
- A macOS 5.1 `STM int64 ref test parallel asymmetric` failed to trigger
- Cygwin tests are still ongoing (2 of them completed part 1 successfully at this point)

In the interest of restoring 4.13 support I'll therefore merge the PR.

PR #347, opened 2023-05-16
Add missing dependency to get 32-bit build running again

### LINKED ISSUE #344: 32-bit pre-setup fails

jmid:
This PR fixes #344 by adding the extra `libgcc-s1` dependency

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -132,7 +132,7 @@ jobs:
           if [ "${COMPILER##*32bit}" = "" ]; then
             sudo dpkg --add-architecture i386
             sudo apt-get update
-            sudo apt-get install pkg-config:i386 libzstd1:i386 libzstd-dev:i386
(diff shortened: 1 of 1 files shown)


jmid:
Lots of (unmodified) targets fail with `exception Invalid_argument("compare: functional value")` which is unrelated to this change to get 32-bit running again, so I'll go ahead and merge.

Issue #351, opened 2023-05-24
Split testing libraries off in a separate repo


jmid:
I'm starting to see more pain points that indicate we should consider splitting off the 3 testing libraries into a separate repo, e.g., `ocaml-multicore/multicoretestlibs`:
- #346 now run an opam installation workflow on all changes - to prevent situations like #342 breaking the newly added 4.13 compatibility. The installation workflow is however also triggered on test suite changes in `src/`. Some of this can be handled by adding conditions on the workflows, but it would be clearer and simpler with separate repos.
- The developer-focused README instructions for running the test suite added in #349 are intermixed with library-user instructions. Both of them would be clearer if separated in each their repo.

These add to the known arguments, such as being able to use an off-the-shelf ocaml-ci on the library-repo.

PR #352, opened 2023-05-31
Support 4.12 too

jmid:
While rolling a new release, opam-ci shows that the `Util` restructuring of #342 makes 4.12 a minimum requirement for the util package. This PR thus makes 4.12 a uniform lower bound for all 3 packages.

It fails the OPAM installation workflow though - due to an unrelated 4.12.1 checksum failure reported here: https://github.com/ocaml/opam-repository/issues/23837
I've checked locally that the packages work on 4.12.1 though.

Again, this is another reason for splitting off the libraries into a separate repo #351, so that this missing lower-bound would have been caught by ocaml-ci.


jmid:
As expected the 4.12 opam install workflow still fails due to the unrelated checksum issue: https://github.com/ocaml-multicore/multicoretests/actions/runs/5131082041/jobs/9230702394

PR #357, opened 2023-06-05
CI: Skip Cygwin part 2 when running only one test

### LINKED ISSUE #350: CI: only_test on Cygwin is run in both halves
I realized that a focus CI run with `only_test` is triggered in both halves of a Cygwin,
for example here https://github.com/ocaml-multicore/multicoretests/actions/runs/5066404458

Ideally, the second half can be skipped in `only_test` mode, which should save some time and CPU cycles.

shym:
The title says it all.
Log of a successful run, where the second part is skipped as intended: https://github.com/shym/multicoretests/actions/runs/5175500758

Closes #350

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -80,9 +80,13 @@ on:
           Defaults to custom_compiler_version chunked at [+~-]
         type: string
         default: ''
+    outputs:
+      skippart2:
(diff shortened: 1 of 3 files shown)


jmid (approved):
LGTM, many thanks! :smiley:

Issue #299, opened 2023-02-20
Lin HashSet tests shrinking taking very long


jmid:
We are experiencing runs of `Lin` `HashSet` tests that take *very* long to shrink.

For example https://github.com/ocaml-multicore/multicoretests/actions/runs/4159533572/jobs/7195669682
on Windows `trunk` took over 4 hours
```
random seed: 401278811
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s Lin DSL Weak HashSet test with Domain
[ ]    0    0    0    0 / 1000     0.0s Lin DSL Weak HashSet test with Domain (generating)
[ ]   15    0    0   15 / 1000    62.0s Lin DSL Weak HashSet test with Domain (shrinking:    5.0003)
[ ]   15    0    0   15 / 1000   122.6s Lin DSL Weak HashSet test with Domain (shrinking:    6.0011)
[ ]   15    0    0   15 / 1000   183.4s Lin DSL Weak HashSet test with Domain (shrinking:   10.0002)
[ ]   15    0    0   15 / 1000   244.7s Lin DSL Weak HashSet test with Domain (shrinking:   10.0016)
[ ]   15    0    0   15 / 1000   306.7s Lin DSL Weak HashSet test with Domain (shrinking:   11.0006)
[ ]   15    0    0   15 / 1000   367.9s Lin DSL Weak HashSet test with Domain (shrinking:   13.0008)
[ ]   15    0    0   15 / 1000   429.4s Lin DSL Weak HashSet test with Domain (shrinking:   14.0005)
[ ]   15    0    0   15 / 1000   490.3s Lin DSL Weak HashSet test with Domain (shrinking:   14.0018)
[...]
[ ]   15    0    0   15 / 1000 14545.9s Lin DSL Weak HashSet test with Domain (shrinking:   90.0189)
[ ]   15    0    0   15 / 1000 14610.8s Lin DSL Weak HashSet test with Domain (shrinking:   90.0201)
[ ]   15    0    0   15 / 1000 14675.2s Lin DSL Weak HashSet test with Domain (shrinking:   90.0214)
[ ]   15    0    0   15 / 1000 14738.1s Lin DSL Weak HashSet test with Domain (shrinking:   90.0226)
[ ]   15    0    0   15 / 1000 14800.6s Lin DSL Weak HashSet test with Domain (shrinking:   90.0238)
[...]

Issue #181, opened 2022-11-11
[ocaml5-issue] Bytecode segfaults on Ephemeron test


jmid:
From #88:

- one bytecode test hit a segfault on the `Ephemeron` test on `trunk` https://github.com/jmid/multicoretests/actions/runs/3444725852/jobs/5747632136#step:8:851

_Originally posted by @jmid in https://github.com/jmid/multicoretests/issues/88#issuecomment-1311983314_
      
- The `Ephemeron` GC-fix was cherry picked to `5.0~trunk` 21 days ago: https://github.com/ocaml/ocaml/pull/11631
- The Effect/bytecode fix was cherry picked 5 days ago: https://github.com/ocaml/ocaml/pull/11704

I'm wondering whether it is a fresh issue - or caused by one of these that somehow haven't propagated through to install via `5.0.0~trunk` and setup-ocaml...

jmid:
The CI just hit another `Ephemeron` segfault on `5.0.0~trunk` - this one however in native code on Linux:
https://github.com/jmid/multicoretests/actions/runs/3448464623/jobs/5755508518
```
random seed: 63947905
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s STM Ephemeron test sequential
[ ]    0    0    0    0 / 1000     0.0s STM Ephemeron test sequential (generating)
[✓] 1000    0    0 1000 / 1000     0.3s STM Ephemeron test sequential

File "src/ephemeron/dune", line 14, characters 0-107:
14 | (rule
15 |  (alias runtest)
16 |  (package multicoretests)
17 |  (deps stm_tests.exe)
18 |  (action (run ./%{deps} --verbose)))
(cd _build/default/src/ephemeron && ./stm_tests.exe --verbose)
Command got signal SEGV.
[ ]    0    0    0    0 / 1000     0.0s STM Ephemeron test parallel
```

For future reference I include a copy of the relevant CI log for the above `Ephemeron` `5.0.0~trunk` bytecode mode segfault - before it is deleted from the GitHub Actions cache:
```
random seed: 382251382
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s STM Ephemeron test sequential
[ ]    0    0    0    0 / 1000     0.0s STM Ephemeron test sequential (generating)
[✓] 1000    0    0 1000 / 1000     1.9s STM Ephemeron test sequential

[...]

Issue #364, opened 2023-06-08
'STM _ ref test parallel asymmetric' failure to trigger


jmid:
Since adding `agree_prop_par_asym` to `STM_domain` in #315 we have seen occasional cases of the `_ ref test parallel asymmetric` failing to trigger.

Here's a fresh case from yesterday, on Cygwin part2 running trunk (5.2.0):
https://github.com/ocaml-multicore/multicoretests/actions/runs/5201990235/jobs/9388312861
```
random seed: 513749813
generated error fail pass / total     time test name
[...]

[ ]    0    0    0    0 / 2000     0.0s STM int ref test parallel asymmetric
[ ] 1501    0    0 1501 / 2000   395.8s STM int ref test parallel asymmetric
[✗] 2000    0    0 2000 / 2000   410.2s STM int ref test parallel asymmetric

[...]
```

Previously this was observed
- in #346 on macOS 5.1 with `int64 ref test parallel asymmetric`
- in #339 on Windows trunk `int ref test parallel asymmetric`
- in #330 on Cygwin trunk and 5.1 with `int ref test parallel asymmetric` and `int64 ref test parallel asymmetric`

It would be nice to use statistics as outlined in #362 to improve the functionality to be more stable.

A fix may involve switching away from a `Semaphore.Binary` as @shym was originally asking about
https://github.com/ocaml-multicore/multicoretests/pull/315#discussion_r1157050399

jmid:
I saw this again on Cygwin part 2 5.1~alpha2
https://github.com/ocaml-multicore/multicoretests/actions/runs/5224077280/jobs/9437485264
```
random seed: 478928519
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s STM int ref test parallel
[ ]    0    0    0    0 / 1000     0.0s STM int ref test parallel (generating)
[ ]    6    0    0    6 / 1000    68.4s STM int ref test parallel (shrinking:    9.0003)
[ ]    6    0    0    6 / 1000   132.7s STM int ref test parallel (shrinking:   11.0002)
[ ]    6    0    0    6 / 1000   193.6s STM int ref test parallel (shrinking:   12.0002)
[ ]    6    0    0    6 / 1000   254.4s STM int ref test parallel (shrinking:   12.0009)
[ ]    6    0    0    6 / 1000   315.6s STM int ref test parallel (shrinking:   13.0008)
[...]

Issue #363, opened 2023-06-08
Lin Thread Queue test reliability


jmid:
I've seen the Lin Thread Queue test find a counterexample under Windows bytecode in less that 250 attempts more than once.

This happened yesterday on 5.1.0~alpha2 windows bytecode:
https://github.com/ocaml-multicore/multicoretests/actions/runs/5199454489/jobs/9376867328
```
random seed: 189508584
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s Lin DSL Queue test with Domain
[ ]    0    0    0    0 / 1000     0.0s Lin DSL Queue test with Domain (generating)
[✓]    2    0    1    1 / 1000     5.1s Lin DSL Queue test with Domain

[ ]    0    0    0    0 /  250     0.0s Lin DSL Queue test with Thread
[✗]   68    0    1   67 /  250    16.8s Lin DSL Queue test with Thread

--- Info -----------------------------------------------------------------------

[...]

--- Failure --------------------------------------------------------------------

Test Lin DSL Queue test with Thread failed (0 shrink steps):

[...]

  Results incompatible with sequential execution

                                               |                    
                           Queue.take t : Error (Stdlib.Queue.Empty)
                                    Queue.take_opt t : None         
                                    Queue.is_empty t : true         
                                    Queue.is_empty t : true         
                                    Queue.is_empty t : true         
                           Queue.take t : Error (Stdlib.Queue.Empty)
                                      Queue.clear t : ()            
                                               |                    
                         .------------------------------------------.
                         |                                          |                    
                Queue.length t : 1                         Queue.add 7 t : ()            
            Queue.take_opt t : Some (7)                Queue.peek_opt t : Some (7)       
                Queue.length t : 1                         Queue.add 1 t : ()            
               Queue.peek t : Ok (4)                       Queue.add 4 t : ()            
               Queue.peek t : Ok (4)                      Queue.peek t : Ok (4)          
              Queue.is_empty t : true                     Queue.peek t : Ok (4)          
     Queue.take t : Error (Stdlib.Queue.Empty)         Queue.take_opt t : Some (4)       
                Queue.clear t : ()                       Queue.take_opt t : None         
                                                           Queue.clear t : ()            

================================================================================
failure (1 tests failed, 0 tests errored, ran 2 tests)
```

And earlier
- on #361 on Windows 5.1 bytecode
- on #349 on Windows bytecode trunk
- on #305 on Windows bytecode trunk

PR #365, opened 2023-06-09
Weak test revisions

jmid:
This PR revises our `Weak` tests...

The `Weak` module depends on the state of the GC.
As such, we have to be extra careful to start from "as clean a starting point as possible".
Just having run sequential tests pollutes the heap state - so better run a `Gc.full_major ()` initially - and a cheaper `Gc.minor ()` in between attempts.

Secondly, the PR switches away from `string`s to more easy-to-shrink `int64`s.

Finally I think running `Lin` tests of `Weak` is misguided, as I don't think it makes sense to perform a fragile GC-dependent parallel run and compare its outputs against more fragile GC-dependent sequential runs.
Parallel `Weak` `STM` tests however work well as stress tests - we previously caught run-time bugs that way (e.g., #181)

Hopefully this will solve #299.

I can see a 5.2 s390x run even completed within the timelimit with this: :tada: 
https://ocaml-multicoretests.ci.dev:8100/job/2023-06-09/150708-ci-ocluster-build-44e4ee


jmid:
The shadowing is handled in 9b85211.
I've also gone for removing the `Lin` tests entirely as I don't see us needing them in the future.
We may have to disable the parallel `src/weak/stm_tests_hashset` test on Windows bytecode where it doesn't seem to consistently trigger.

jmid:
CI summary for 720028c
- a failure to trigger a parallel `weak/stm_tests_hashset` on Windows bytecode 5.1.0~alpha2
- a timeout due to excessive `Lin` `Ephemeron` shrinking on Cygwin part 1 5.1.0~alpha2
- a timeout due to excessive `Lin` `Out_channel` shrinking on Windows bytecode trunk 5.2
- a timeout due to `threadomain` livelock on Windows 5.1.0~alpha2 and on Windows trunk 5.2 #203 
- a timeout due to a on `domain_spawntree with Atomic` on Windows trunk 5.2 #354
- a timeout due to long-running `int ref STM` and `Thread` tests (several 1500+ secs) on Cygwin part 2 trunk 5.2
- a failure to trigger a `Lin` `Out_channel` test on Linux arm64 5.1 #360

Modulo the first one, none of this is related to this PR.

shym:
Small suggestion for the last commit: rather than disabling the warning globally, it can be disabled only where it might trigger, it’s easier to understand why it is required.
```diff
@@ -41,6 +41,7 @@ struct
 
   module Int64 =
   struct
+    [@@@warning "-unused-value-declaration"]
     (* support Int64.hash added in 5.1 *)
     external seeded_hash_param :
       int -> int -> int -> 'a -> int = "caml_hash" [@@noalloc]
```

jmid:
CI summary for the latest run:
- 1 Windows bytecode `5.1.0~alpha2` `threadomain` timeout and 1 crash - with a new variant #203
- 1 Windows bytecode trunk 5.2 unexpected `Thread` `Queue` counterexample #363 
- 2 timeouts on Cygwin part 1 `5.1` and 1 Linux s390x `5.1` due to several (primary `Lin`) tests taking 1000+ seconds
- 1 macOS arm64 `5.2` failure to trigger `STM int ref test parallel asymmetric` #364 

I'll go ahead and merge.

PR #366, opened 2023-06-12
Teach `dune` to focus on one particular test

shym:
This PR adds the possibility to use `dune` to focus on one particular test, in a similar way to what we do with github actions. This will make it easy to enable focus testing in our CI by using `dune build @@runtest ...` rather than `dune runtest ...`.

The capability was tested with CI on commit [a247a5e50f](https://github.com/shym/multicoretests/tree/a247a5e50f532206b719e2adec014ee7e11be48c) (thanks for the OPAM workflows which caught early the missing `package` annotation! :+1:).

We could imagine using that same mechanism to run focused tests also in github actions:
- Pros: this would use the same configuration across testing platforms,
- Cons: this would introduce `bash` in the testing stack on Windows / Cygwin (I don’t know whether this could introduce noise or not); this would probably require setting some CI parameter to avoid double runs on Cygwin (which was patched in #357).

While doing this, I wondered whether it would be useful to split the test suite into more aliases than the single `runtest`, for instance to separate some tests that are prone to sporadic failures :thinking:

### DIFF
--- a/dune
+++ b/dune
@@ -7,3 +7,93 @@
  (name default)
  (package multicoretests)
  (deps (alias src/default)))
+
+; The aliases to control what is run in CI
+; It can either be the full test suite, or focus on a single test
+; Aliases ci1 and ci2 are useful to split the test suite into
+; subparts, for platforms such as Cygwin which are too slow
+(alias
+ (name ci)
+ (package multicoretests)
+ (deps
+  (alias_rec %{env:DUNE_CI_ALIAS=runtest})))
+  ; (alias_rec focusedtest)))
+
+(alias
+ (name ci1)
+ (package multicoretests)
+ (deps
+  (alias_rec src/array/runtest)
+  (alias_rec src/atomic/runtest)
+  (alias_rec src/bigarray/runtest)
+  (alias_rec src/buffer/runtest)
+  (alias_rec src/bytes/runtest)
+  (alias_rec src/domain/runtest)
+  (alias_rec src/dynlink/runtest)
+  (alias_rec src/ephemeron/runtest)
+  (alias_rec src/floatarray/runtest)
+  (alias_rec src/hashtbl/runtest)
+  (alias_rec src/io/runtest)))
+
+(alias
+ (name ci2)
+ (package multicoretests)
+ (deps
+  (alias_rec src/lazy/runtest)
+  (alias_rec src/neg_tests/runtest)
+  (alias_rec src/queue/runtest)
+  (alias_rec src/semaphore/runtest)
+  (alias_rec src/stack/runtest)
+  (alias_rec src/sys/runtest)
+  (alias_rec src/thread/runtest)
+  (alias_rec src/threadomain/runtest)
+  (alias_rec src/weak/runtest)))
+
+; @focusedtest
+; repeat a single test a couple of times
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; A rule to repeat the test executable given as dependency a couple of
+; times and report at the end whether this worked
+
+; To change the test to repeat, change the source of the `copy`:
+(rule
+ (copy src/io/lin_tests_dsl_domain.exe focusedtest.exe))
+
+(rule
+ (alias focusedtest)
+ (package multicoretests)
+ (deps focusedtest.exe)
+ (enabled_if (<> %{os_type} Win32))
+ (action
+  (no-infer
+   (progn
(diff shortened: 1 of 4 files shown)


shym:
> - Is there something special about the `@@runtest` alias? `dune runtest` is a shorthand for `dune build @runtest` - but I'm too dune-illiterate to know `@@runtest`... :thinking: 

`@runtest` is `(alias_rec runtest)` (so `runtest` is all subdirectories) whereas `@@runtest` is `(alias runtest)` (only in the current directory).

> - [...] using a (not-so-pretty) generated dune file, akin to what I did here [...]

I don’t understand how this works in that example, but that’s an idea. How useful do you find the feature to run all the tests and get a final count of the number of failures at the end? (I didn’t find how to get this without `bash`, but it might be possible with `cmd`?).

shym:
I rewrote that PR last week to take your comment into account. It is now providing:

- a new `@ci` alias to be used in CI (and configured to depend on what we want to run), instead of modifying the `@runtest` alias, since the goal is more explicit:
  - by default, `@ci` will build the alias in the `$DUNE_CI_ALIAS` environment variable if it exists, or `@runtest`,
- support for Windows to focus on one test, using `cmd` rather than `bash` to repeat a test,
- more of the logic for focused tests is converted to dune action, using `diff` to error out if at least one of the runs failed (listing the iterations that failed, by the way, so it will be faster to find them in the log),
- for slow platforms (aka Cygwin), two aliases `@ci1` and `@ci2` are provided, to split up the test suite into manageable pieces,
- the Github `common` workflow is greatly simplified.

The last commit on that [branch](https://github.com/shym/multicoretests/tree/dune-onetest-2) shows how to focus on one test (and modify the repeat counts).

dune
```diff
@@ -7,3 +7,93 @@
...
+  (no-infer
+   (progn
+    (system "echo skipnextjob=true >> %{env:GITHUB_OUTPUT=dummy_file}")
+     ; ^ this is to ensure only one job is run in multi-job GitHub
+     ; workflows (only Cygwin at the moment)
+
+    (write-file hoped "")
+    (write-file failed-runs "")
+    (run cmd /q /c
+     "for %G in (1,2,3,4, 5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20) do (echo Starting %G-th run && focusedtest.exe -v || echo %G >> failed-runs)")
```
jmid, line 97: Nit/OCD trigger warning: extra space after `4` in `1,2,3,4, 5,6,...`

jmid:
Cool! I think I follow :sweat_smile: 
You're using the environment variable `DUNE_CI_ALIAS` only to support breaking up the Cygwin workflow in two IIUC?
That means it can be eliminated once we identify the reason for why Cygwin is taking so long and get back to running it in one workflow (which would again enable further `dune` simplifications...)

With this, if we switch Multicoretests-CI to run `opam exec -- dune build "@ci" -j1 --no-buffer ...` too we would enable 'focused tests' there too, right?

What happens if one runs `dune runtest ...` in the console on the PR's setup? :thinking: 
Will it perform a full run as usual, while the CI system's `@ci` alias allow either a full run or a focused test run?

shym:
> You're using the environment variable `DUNE_CI_ALIAS` only to support breaking up the Cygwin workflow in two IIUC?

That’s its only use, at least as it stands. And as it is easy to override the environment variable only in Github CI, I have no other application in mind.

> With this, if we switch Multicoretests-CI to run `opam exec -- dune build "@ci" -j1 --no-buffer ...` too we would enable 'focused tests' there too, right?

Exactly!

> What happens if one runs `dune runtest ...` in the console on the PR's setup? thinking Will it perform a full run as usual, while the CI system's `@ci` alias allow either a full run or a focused test run?

The standard `runtest` alias is left unmodified, so no strange effect there.

jmid:
> The standard runtest alias is left unmodified, so no strange effect there.

That's a beautiful solution! :heart: 
We get both the benefit of optional 'focused tests' in the CI without affecting the "usual console workflow" :tada:

jmid (approved):

jmid:
CI summary
- macOS 5.1 failure to trigger both `STM {int,int64} ref test parallel asymmetric`
- Cygwin 5.1 and trunk part 1 cancelled(?!)
- Linux-arm64-5.1 and Linux-arm64-5.2 failure to trigger `Lin DSL Out_channel test with Domain`
- Linux-s390x-5.2 lots of failures during `thread_createtree - with Atomic` with error message
  `Thread 1175477 killed on uncaught exception Sys_error("Thread.create: Resource temporarily unavailable")`

None of these are caused by the changes of the current PR.
The last one is however interesting as we haven't seen it before IIRC... :thinking:

Issue #373, opened 2023-07-10
Long In_channel shrink runs causing timeout


jmid:
Whereas #370 brought improvements, it also triggers a new issue in the form of very long shrink sequences, to the point of them causing a timeout.

Here's one from 5.1 32-bit (i.e. bytecode):
https://github.com/ocaml-multicore/multicoretests/actions/runs/5488638507/jobs/10001760322
```
random seed: 394954226
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s Lin DSL In_channel test with Domain
[ ]    0    0    0    0 / 1000     0.0s Lin DSL In_channel test with Domain (generating)
[ ]   67    0    0   67 / 1000    60.1s Lin DSL In_channel test with Domain (shrinking:   37.0006)
[ ]   67    0    0   67 / 1000   120.3s Lin DSL In_channel test with Domain (shrinking:   67.0006)
[ ]   67    0    0   67 / 1000   180.4s Lin DSL In_channel test with Domain (shrinking:   96.0010)
[ ]   67    0    0   67 / 1000   240.5s Lin DSL In_channel test with Domain (shrinking:  129.0005)
[ ]   67    0    0   67 / 1000   300.6s Lin DSL In_channel test with Domain (shrinking:  157.0003)
[ ]   67    0    0   67 / 1000   361.1s Lin DSL In_channel test with Domain (shrinking:  182.0008)
[ ]   67    0    0   67 / 1000   421.2s Lin DSL In_channel test with Domain (shrinking:  208.0008)
[ ]   67    0    0   67 / 1000   481.6s Lin DSL In_channel test with Domain (shrinking:  233.0006)
[ ]   67    0    0   67 / 1000   541.7s Lin DSL In_channel test with Domain (shrinking:  258.0008)
[ ]   67    0    0   67 / 1000   601.8s Lin DSL In_channel test with Domain (shrinking:  289.0009)
...
[ ]   67    0    0   67 / 1000 10589.2s Lin DSL In_channel test with Domain (shrinking: 4304.0004)
[ ]   67    0    0   67 / 1000 10649.5s Lin DSL In_channel test with Domain (shrinking: 4321.0010)
[ ]   67    0    0   67 / 1000 10709.6s Lin DSL In_channel test with Domain (shrinking: 4340.0009)
[ ]   67    0    0   67 / 1000 10770.0s Lin DSL In_channel test with Domain (shrinking: 4360.0004)
[...]

PR #375, opened 2023-07-13
`In_channel`: Generate and shrink bytes only on length

shym:
This PR spawned out of #373.

In `In_channel`, `bytes` are used as input buffers so only their lengths is relevant: generate and shrink them by taking only their lengths into account

It currently contains a focusing commit to remove before merging, so I set this PR as a draft.

### DIFF
--- a/src/io/lin_tests_spec_io.ml
+++ b/src/io/lin_tests_spec_io.ml
@@ -9,7 +9,15 @@ module ICConf : Lin.Spec = struct
   let cleanup = In_channel.close
 
   open Lin
-  let int,int64,bytes = nat_small,nat64_small,bytes_small
+  let int,int64 = nat_small,nat64_small
+
+  let bytes =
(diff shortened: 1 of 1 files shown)


jmid:
I rebased to drop the two focusing commits and force-pushed the result.
I'll let the CI finish and then plan to merge.

jmid:
CI summary:
- Cygwin trunk + Windows 5.1 timeouts during to excessive (9900.4s + 9729.7s) `Out_channel` shrink runs #378
- Windows bytecode 5.1 + Windows 5.2/trunk timeout on `threadomain` #203
- ppc64le 5.2/trunk crashes on `STM Array/Bytes test sequential` #380
- 32-bit trunk unexpectedly found a `Lin DSL Queue test with Thread` counterexample #363

Since there were no `In_channel` timeouts in sight I'll merge.

PR #376, opened 2023-07-14
Thread joingraph unification (chore)

jmid:
The PR rewrites the two separate `Atomic` and `tak` tests inside `src/thread/thread_joingraph` into one.
They were written, quick-and-dirty, over 1 year ago and large left untouched (see the git history).

This is a step toward bringing them closer to the style of the effective `src/threadomain` test.
As part of the rewrite it now occasionally triggers an explicit `Gc.minor` call.

The rewrite will make it easier to later experiment with and add additional functionality (signal raising, finalizers, ...)
If nothing else, it will also shave a couple of seconds of the test suite run time.

### DIFF
--- a/src/thread/thread_joingraph.ml
+++ b/src/thread/thread_joingraph.ml
@@ -7,7 +7,8 @@
 open QCheck
 
 (* Generates a sparse DAG of join dependencies                        *)
-(* Each thread is represented by an array index w/at most 1 dep. each *)
+(* Each thread is represented by record with an optional array index
+   to model at most 1 dependency each *)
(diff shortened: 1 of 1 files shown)


jmid:
CI summary:
- Cygwin trunk/5.2 part1 - timeout during a 4528.5s `Lin DSL Out_channel test with Domain` shrinking run
- Windows 5.1 - timeout due to a 11331.5s `Lin DSL Out_channel test with Domain` shrinking run
- 2 x Windows trunk/5.2 - timeout during `threadomain`

I'll go ahead and merge

Issue #343, opened 2023-05-12
Remove duplicated code for value pretty-printers


shym:
`Lin` and `STM` both use `'a -> string` functions in their main type to display values but they each redefine their set of functions for common types. All this machinery should be shared and so moved to `Util`.
`Util.Pp` introduced in #342 would be a good candidate for that.
Beside sharing the code, the `'a -> string` type could be replaced with `'a Util.Pp.t` to get prettier outputs (ie with parentheses when needed) if we deem it not to break too much user code.

Issue #308, opened 2023-03-09
Abbreviating print functions


jmid:
Currently we are printing `string`s, `list`s, etc. unmodified which can sometimes get a bit long to the point that it may detract from seeing the structure of the program causing an issue. Below I list two example outputs that seem a bit excessive.

In contrast, the OCaml top-level cuts the output off at some point:
```
# String.make 4000 'c';;
- : string =
"ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc"... (* string length 4000; truncated *)
```
We may want to have our `string`, `list`, ... printing functions do something similar?


Here is an example with a `string`
```
Messages for test Lin DSL In_channel test with Domain:

  Results incompatible with sequential execution

                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
[...]

PR #379, opened 2023-07-28
Improve `Util.Pp` pretty-printers

shym:
This PR brings a few features:
- usage of `Format` boxes to provide the same formatting as `deriving show` _and_ abide by `Format`’s rule that all content should be output into some box,
- the possibility to require the output to be truncated,
- a couple of pretty-printers that were missing,
- a couple of typo fixes.

The possibility to truncate would address #308 if all pretty-printers in `STM` and `Lin` were always using `Util.Pp` (this is #343).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Next
 
+- #379: Extend the set of `Util.Pp` pretty-printers and teach them to
+  add break hints similar to `ppx_deriving.show`; teach `to_show` to
+  generate truncated strings when `$MCTUTILS_TRUNCATE` environment
+  variable is set
 - #368: Switch `STM_domain.agree_prop_par_asym` from using
   `Semaphore.Binary` to using an `int Atomic.t` which improves
   the error rate across platforms and backends
--- a/dune
+++ b/dune
@@ -1,5 +1,11 @@
 (env
- (debug-runtime (link_flags :standard -runtime-variant=d))
+ (debug-runtime
+  (link_flags :standard -runtime-variant=d)
+  (env-vars
+   (MCTUTILS_TRUNCATE 50)))
+ (_
+  (env-vars
+   (MCTUTILS_TRUNCATE 50)))
 )
 
 ;; make `dune build` target a recursive default target
--- a/lib/util.ml
+++ b/lib/util.ml
@@ -127,20 +127,22 @@ module Pp = struct
         s
 
   let of_show f par fmt x =
-    fprintf fmt (if par then "(%s)" else "%s") (f x)
+    fprintf fmt (if par then "@[(%s)@]" else "@[%s@]") (f x)
 
   let cst0 name fmt = pp_print_string fmt name
 
   let cst1 (pp : 'a t) name par fmt x =
-    fprintf fmt (if par then "(%s %a)" else "%s %a") name (pp true) x
+    let o, c = if par then ("(", ")") else ("", "") in
+    fprintf fmt "%s@[<2>%s@ %a@]%s" o name (pp true) x c
 
   let cst2 (pp1 : 'a t) (pp2 : 'b t) name par fmt x y =
-    fprintf fmt (if par then "(%s (%a, %a))" else "%s (%a, %a)") name (pp1 false) x (pp2 false) y
+    let o, c = if par then ("(", ")") else ("", "") in
+    fprintf fmt "%s@[<2>%s (@,%a,@ %a)@]%s" o name (pp1 false) x (pp2 false) y c
 
   let cst3 (pp1 : 'a t) (pp2 : 'b t) (pp3 : 'c t) name par fmt x y z =
-    fprintf fmt
-      (if par then "(%s (%a, %a, %a))" else "%s (%a, %a, %a)")
-      name (pp1 false) x (pp2 false) y (pp3 false) z
+    let o, c = if par then ("(", ")") else ("", "") in
+    fprintf fmt "%s@[<2>%s (@,%a,@ %a,@ %a)@]%s" o name (pp1 false) x
+      (pp2 false) y (pp3 false) z c
 
   let pp_exn = of_show Printexc.to_string
   let pp_unit _ fmt () = pp_print_string fmt "()"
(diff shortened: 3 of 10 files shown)


jmid:
LGTM, thanks!
I think an environment variable is the right solution, to allow for both a "shortened print" and a "full one" without having to recompile.

Should we do something similar for `bytes`?
Would it make sense to add unit/expect tests to ensure that it works as intended and protect ourselves against future breakage? :thinking: 
Finally (for another PR) I agree that we should seriously consider switching Lin+STM to use these for consistent 
argument and result printing.

shym:
I’ve added an internal test for the pretty-printers, with variants for the truncation.

> Should we do something similar for bytes?
I don’t see what you have in mind, `Util.Pp` contains a `pp_bytes` pretty-printer.

jmid:
> I don’t see what you have in mind, `Util.Pp` contains a `pp_bytes` pretty-printer.

My bad, I was too fast reading the cut off code as only applying to strings and not bytes :facepalm: 
Since this is located in `to_show` I assume the cut off applies to all to-string-coercions (string, bytes, list, ...) passing through that function, right?

jmid:
The added tests look great BTW, and the expect outputs nicely answer the above question :sweat_smile: 

I guess having it print the final closing quote or square bracket like the OCaml toplevel will require some form of type-directed approach?
```ocaml
utop # String.make 375 'a';;
- : string =
"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"... (* string length 375; truncated *)
utop # List.init 375 (fun _ -> [0]);;
- : int list list =
[[0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; 
 [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [0]; [...]; ...]
```

This is just a thought - not meant as critique. I'm quite happy with this as a first version!

shym:
> The added tests look great BTW, and the expect outputs nicely answer the above question 😅

:sweat_smile:

> I guess having it print the final closing quote or square bracket like the OCaml toplevel will require some form of type-directed approach?

I think so. At the level where the truncating mechanism is located in that PR, it sees only the stream of bytes to output, without structure. I wondered whether the boxing or tagging mechanisms in formatters could be retargeted for that kind of purposes, but I didn’t find how.

shym:
Reviewing the CI logs I realise (finally!) that (obviously, as it is one of the main goal of `Format`) the result produced by `to_show` can be a multiline string while `print_triple_vertical` assumes it is not the case (see this [example](https://github.com/ocaml-multicore/multicoretests/actions/runs/5942185755/job/16114588868?pr=379#step:18:239) for instance).
(It must have been the case already with the pretty-printers generated by `ppx_deriving.show` since we have something really similar, hidden by the fact that long lines are not readable anyway)

A simple way to fix pretty-printing with `print_triple_vertical` is to set the formatter margin to `max_int`.
But now I wonder: do we want `to_show` functions to always generate one-liners? Or should we add the margin as an (optional?) argument to override the default 78? :thinking:

jmid:
First of all, let me state that `print_triple_vertical` is an ugly piece of code that I'm not proud of! :sweat_smile: 
(I don't know whether it can be written more cleanly with Formatters).
I agree that it assumes 1-line `cmd` strings when stitching them together.

Code aside, too long outputs (strings, bytes, lists, ...) mangle the upside down Y-shape printed, which motivated the request for abbreviating print functions in #308

In CI logs, I'd rather skim over, e.g., an abbreviated string in an comprehendable Y-shape rather than see a long string in full detail. That suggests we should set the environment variable in the CI workflows.
I suspect other end-users would prefer such a default setup too :thinking: 

It would however be nice to be able to print the full string some way (locally and in the CI), say to create a standalone reproducer. That should however be possible by resetting the `MCTUTILS_TRUNCATE` environment variable.

To keep things nice and without wrapped lines in the former scenario, setting the formatter margin to `max_int` sounds reasonable.

There may be cases I'm overlooking, say record types with lists in them that could warrant multi-line formatted strings. So far, such types haven't occurred in the interfaces we have tested :shrug:

jmid:
CI summary for the latest run 64c1902
- Cygwin trunk part 2 failed to trigger `STM Sys test parallel`
  before also timing out during a 253.7s `Lin CList int64 test with Thread`
  after a 2403.4s `STM Sys test parallel`, 1071.1s `STM Weak test parallel`, 3267.4s `STM Weak HashSet test parallel`, 5028.3s `Lin ref int64 test with Thread` and 1424.6s `Lin CList int test with Thread` #384 
- Linux ppc64le failed with an `STM Bytes test sequential` segfault and an `STM Float Array test sequential` model difference #380

Neither of these are related to the PR, so I'll go ahead and merge

jmid:
Actually, it would be nice with a CHANGES entry to document changes in the user-facing Util.Pp module.
Could I convince you to add one @shym?

shym:
Indeed!
Should I add also a default setting to `MCTUTILS_TRUNCATE` to 50 I’ve tested locally? As `Util.Pp` are used only for STM commands, it won’t bring the hoped truncation for the worse offenders (`Dynlink`, say) just yet, though.

jmid:
> Should I add also a default setting to `MCTUTILS_TRUNCATE` to 50 I’ve tested locally? 

Yep, please do! :pray:

jmid:
CI summary for 13d87f8
- Win bytecode trunk/5.2 timeout during `threadomain` #203
- Cygwin trunk/5.2 timeout during 1343.7s `Lin CList int test with Thread` after 2276.9s `STM Sys test parallel`, 10291.7s `Lin ref int64 test with Thread` #384
- Linux ppc64le trunk/5.2 failure with a `STM Array test sequential` `ABRT` and a `STM Float Array test sequential` `SEGV` #380

None of these are related to the PR, so I'll merge (and I mean it this time!)

PR #392, opened 2023-09-07
Add Lin Domain.DLS test

jmid:
This PR adds a quick Lin stress test of `Domain.DLS`.

It uses only one `DLS.key` for a start - not impressive but it is a start.
We might revisit it based on `multiple-ts` or consider replacing it by an `STM` test later.

Without the optional `~split_from_parent` the child domain does not inherit the key, which makes for a fast negative test.
I did notice that the test takes longer and longer as it progresses, which may be due to the inability to remove `DSL.key`s again. For this reason, it seems faster to run the negative test first... 
We might consider (a) running the property/test in child domain or (b) splitting the two tests to run in separate processes.

### DIFF
--- a/src/domain/dune
+++ b/src/domain/dune
@@ -17,3 +17,11 @@
  (libraries util qcheck-core qcheck-core.runner)
  (action (run %{test} --verbose))
 )
+
+(test
+ (name lin_tests_dsl_dls)
+ (modules lin_tests_dsl_dls)
+ (package multicoretests)
+ (libraries qcheck-lin.domain)
+ (action (run %{test} --verbose))
+)
--- a/src/domain/lin_tests_dsl_dls.ml
+++ b/src/domain/lin_tests_dsl_dls.ml
@@ -1,0 +1,25 @@
+open Domain
+open Lin
+
+module DLSConf = struct
+  type t = int DLS.key
+  let init () = DLS.new_key (fun () -> 0) (* without split from parent *)
+  let cleanup _ = ()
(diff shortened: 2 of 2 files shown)


shym (approved):
Nice little test!
As keys cannot be freed and as `new_key` is really only called from `init`, it might make sense to use a constant key? :thinking: Or allocate a list of keys initially and pick one of them in `init`?

jmid:
I grew increasingly happy with the short-comings the initial version, and therefore set out to write an `STM` test addressing some of them in ef4b42d. 

The resulting `STM` test
- runs the two test properties in a child `Domain` (to avoid `Domain.DLS` polution across repetitions) and
- allocates an initial list of keys as suggested

I also adjusted the `Util.repeat` and `retries` parameters as a lot repetition didn't seem necessary to trigger it, and since I needed to write out the properties anyway to run the test properties in a child `Domain`.

This results in a negative test, that is quickly able to produce a counterexample where simultanious changes in two children are not seen in the sibling:
```

                               |                           
                               |                           
            .------------------------------------.
            |                                    |                           
     Set (2, 1) : ()                      Set (0, 43) : ()                   
        Get 0 : 0                            Get 2 : 2            
```

jmid:
A round of focused tests confirmed that the `STM Domain.DLS` test runs fast and works reasonable well.
I've therefore disabled the initial `Lin` test.

I expect this full CI run to fail its `trunk` workflows without #12545 merged into ocaml/ocaml.

Issue #354, opened 2023-06-02
[ocaml5-issue] Windows trunk bytecode domain_spawntree crash or deadlock


jmid:
Today surfaced a Windows trunk bytecode crash on src/domain/domain_spawntree.ml
https://github.com/ocaml-multicore/multicoretests/actions/runs/5154525696/jobs/9283085877
```
random seed: 502502158
generated error fail pass / total     time test name

[ ]    0    0    0    0 /  100     0.0s domain_spawntree - with Atomic
File "src/domain/dune", line 14, characters 7-23:
14 |  (name domain_spawntree)
            ^^^^^^^^^^^^^^^^
(cd _build/default/src/domain && ./domain_spawntree.exe --verbose)
Command exited with code -1073741819.
[ ]    0    0    0    0 /  100     0.0s domain_spawntree - with Atomic (generating)
```

jmid:
Found another occurrence of this causing a live/deadlock:
https://github.com/ocaml-multicore/multicoretests/actions/runs/5242663626/jobs/9466351902
```
random seed: 320533040
generated error fail pass / total     time test name

[ ]    0    0    0    0 /  100     0.0s domain_spawntree - with Atomic
[ ]    0    0    0    0 /  100     0.0s domain_spawntree - with Atomic (generating)Terminate batch job (Y/N)? 
^CFatal error: exception User interruption
Error: The operation was canceled.
```

jmid:
Observed another variant of this on the Mingw Windows 5.0.0 workflow
https://github.com/ocaml-multicore/multicoretests/actions/runs/5565429087/job/15072781545
```
random seed: 221745155
generated error fail pass / total     time test name

[ ]    0    0    0    0 /  100     0.0s domain_spawntree - with Atomic
Fatal error: no domain lock held
File "src/domain/dune", line 14, characters 7-23:
14 |  (name domain_spawntree)
            ^^^^^^^^^^^^^^^^
(cd _build/default/src/domain && ./domain_spawntree.exe --verbose)
Command exited with code 3.
[ ]    0    0    0    0 /  100     0.0s domain_spawntree - with Atomic (generating)
```

jmid:
Crash seen again on Mingw bytecode trunk:
https://github.com/ocaml-multicore/multicoretests/actions/runs/6093487146/job/16533243147
```
random seed: 373304996
generated error fail pass / total     time test name

[...]

PR #395, opened 2023-09-12
Add latest issues to README

jmid:
This little PR adds the latest 4 acknowledged issues to the README.

While I was at it,
- I also fixed an older typo and
- toned down the occurrences of  'young', 'recent' and 'OCaml 5.0'

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Multicore tests
 Property-based tests of (parts of) the OCaml multicore compiler and run time.
 
 This project contains
-- a randomized test suite of OCaml 5.0, packaged up in `multicoretests.opam`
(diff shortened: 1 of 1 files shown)


jmid:
CI summary
- Mingw 5.1 timeout in `threadomain` #203
- Mingw 5.1 timeout in `domain_spawntree - with Atomic` #354
- Llinux-s390x-5.2 failed due to already reported cmi lookup regression 
- Linux-ppc64le-5.2 failed due to already reported cmi lookup regression 
 - 7 runs failed within 1 min due to network issues

Out of 50 + 9 runs, we thus had 10 faiures, 8 were for outside reasons and 2 were genuine failures (no false positives!)
I'll merge.

Issue #446, opened 2024-03-26
[ocaml5-issue] Segfault in `STM Domain.DLS test sequential` on bytecode


jmid:
In the CI-run for #445 on 32-bit trunk the `STM Domain.DLS test sequential` triggered a segfault
https://github.com/ocaml-multicore/multicoretests/actions/runs/8436771284/job/23104952265?pr=445
```
random seed: 107236932
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s STM Domain.DLS test sequential
File "src/domain/dune", line 31, characters 7-20:
31 |  (name stm_tests_dls)
            ^^^^^^^^^^^^^
(cd _build/default/src/domain && ./stm_tests_dls.exe --verbose)
Command got signal SEGV.
[ ]    0    0    0    0 / 1000     0.0s STM Domain.DLS test sequential (generating)
```

This may be another case of a 32-bit/bytecode issue showing up in a couple of different tests:
- #440 
- #412 

Surprisingly this case however triggered in a sequential (single-domain) test! :open_mouth:

gasche:
I wonder if it is related to #12889, the only recent change to Domain.DLS that I can think of. (I hope not!)

jmid:
This just triggered again on 32-bit `5.3.0+trunk` by the merge to `main` of #460:
https://github.com/ocaml-multicore/multicoretests/actions/runs/9169655398/job/25210472949
```
random seed: 103830913
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s STM Domain.DLS test sequential
File "src/domain/dune", line 31, characters 7-20:
31 |  (name stm_tests_dls)
            ^^^^^^^^^^^^^
(cd _build/default/src/domain && ./stm_tests_dls.exe --verbose)
Command got signal SEGV.
[ ]    0    0    0    0 / 1000     0.0s STM Domain.DLS test sequential (generating)
```

gasche:
Is there more information that we can use to try to investigate this? "There is a segfault somewhere in Domain.DLS on 32bit" is not that much.

jmid:
First off, this is a collection of failures we observe. 
Once we have fleshed out reproducible steps, these are reported upstream.
Help is very welcome, snarky remarks less so.

> "There is a segfault somewhere in Domain.DLS on 32bit" is not that much.

[...]

Issue #440, opened 2024-03-05
[ocaml5-issue] segfault in 32-bit mode during neg_tests/lin_tests_domain.ml


jmid:
The merge to main of #438 crashed with a segfault on trunk in 32-bit mode:
https://github.com/ocaml-multicore/multicoretests/actions/runs/8156987450/job/22295610813
```
random seed: 86271143
generated error  fail  pass / total     time test name

[ ]     0     0     0     0 / 10000     0.0s Lin ref int test with Domain
[ ]     0     0     0     0 / 10000     0.0s Lin ref int test with Domain (generating)
[✓]     1     0     1     0 / 10000     3.7s Lin ref int test with Domain

[ ]     0     0     0     0 / 10000     0.0s Lin ref int64 test with Domain
[✓]     1     0     1     0 / 10000     1.0s Lin ref int64 test with Domain

[ ]     0     0     0     0 / 10000     0.0s Lin CList int test with Domain
[✓]     3     0     1     2 / 10000     5.0s Lin CList int test with Domain

[ ]     0     0     0     0 / 10000     0.0s Lin CList int64 test with Domain
[✓]     3     0     1     2 / 10000     1.8s Lin CList int64 test with Domain

--- Info -----------------------------------------------------------------------

Negative test Lin ref int test with Domain failed as expected (17 shrink steps):

                            |            
                            |            
                 .---------------------.
                 |                     |            
          Sut_int.decr t        Sut_int.set t 1     
                                 Sut_int.get t      


+++ Messages ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

Messages for test Lin ref int test with Domain:

File "src/neg_tests/dune", line 73, characters 7-23:
73 |  (name lin_tests_domain)
            ^^^^^^^^^^^^^^^^
(cd _build/default/src/neg_tests && ./lin_tests_domain.exe --verbose)
Command got signal SEGV.
```

jmid:
Fixed in https://github.com/ocaml/ocaml/pull/13553

Issue #412, opened 2023-11-12
[ocaml5-issue] Segfault in Lin Out_channel test in 32-bit and bytecode mode


jmid:
On #411 we saw a segfault in `Lin Out_channel test with Domain` in a 32-bit trunk/5.2 run:
https://github.com/ocaml-multicore/multicoretests/actions/runs/6826699760/job/18567181221?pr=411
```
random seed: 125264582
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s Lin In_channel test with Domain
[ ]    0    0    0    0 / 1000     0.0s Lin In_channel test with Domain (generating)
[✓]    4    0    1    3 / 1000    12.8s Lin In_channel test with Domain

File "src/io/dune", line 21, characters 7-23:
21 |  (name lin_tests_domain)
            ^^^^^^^^^^^^^^^^
(cd _build/default/src/io && ./lin_tests_domain.exe --verbose)
Command got signal SEGV.
[ ]    0    0    0    0 / 5000     0.0s Lin Out_channel test with Domain
```

PR #448, opened 2024-03-27
Add debug bytecode

jmid:
Motivated by recent crashes in bytecode:
- #412
- #440 
- #446

this WIP PR adds bytecode debug workflows for Linux, macOS, and MinGW.


jmid:
Rebased and updated
- to reflect the macOS ARM64 vs Intel workflow split
- to add scheduled runs on these new workflows

jmid:
Rebased and updated
- to reflect the new GitHub actions setup and
- adding a 5.4.0 workflow too

This is now limited to just Linux bytecode debug. We have seen mostly runtime assertion failures recently, so it should be helpful to catch these under the bytecode interpreter too.

jmid:
Note to self: Like the native debug workflows, this reuses the cache entries for the non-debug workflows, as both debug and non-debug can be run from the same compiler build. Here are example keys:
- `ocaml-a8e2e13a317facf1cfb9c973dd211ea11b192c1a-linux-x86_64-bytecode-only`
- `ocaml-bc5083c1d3b773fc3198355494afd2fb4628ff0e-linux-x86_64-bytecode-only`
- `ocaml-f28bfbcb232b017460cb0ba7f9ec8cf1cf773811-linux-x86_64-bytecode-only`

jmid:
FTR, these are the times for the bytecode debug workflows:
- Bytecode 5.2 debug / Successful in 180m
- Bytecode 5.3 debug / Successful in 182m
- Bytecode trunk debug / Successful in 194m

It seems saying - and excessive - that the fastest ones of those considered take 3h or more.
I'll therefore close this for now.

PR #454, opened 2024-04-18
Omission: temporarily disable Dynlink test under Cygwin too

jmid:

PR #455, opened 2024-04-19
Increment count for STM int64 ref test with Thread

jmid:
This failed to trigger a counterexample on macos-arm64-5.2 within 2500 iterations on the merge of #454 to `main`.


jmid:
CI summary
- Cygwin trunk timed out during `Lin.Internal CList int test with Thread` after spending `1449.1s Lin Bytes test with Domain` and `6330.6s Lin Hashtbl test with Domain`

The latter seems rather excessive! 

Out of 37 workflows with a borderline genuine issue / false alarm

PR #456, opened 2024-04-20
Specifications can be terser through more type trickery

OlivierNicole:
As I was writing an STM specification for Dynarrays, I was a bit frustrated to have to add a catch-all `assert false` at the end of `postcond`. So I went into an over-engineering rabbit hole and found that it can be avoided by making the `cmd` type of specifications a GADT, which allows to specify the result type for commands. The typechecker is then able to deduce the result type when pattern matching on commands.

As an example, I have rewritten `src/atomic/stm_tests.ml` to match the new `Spec` signature.

This has a number of benefits and some drawbacks.

## Benefits

- When writing the `run` function, results no longer need to be wrappend into the `Res` constructor. However, it is still necessary to also return an instance of `'a ty_show`; I have not found a simpler way to make the results printable.
- When writing the `postcond` function, it is no longer necessary to match on patterns like `Res ((Result (Int, Exn), _), v)` for the typechecker to deduce the right type for `v`: the command name suffices. It is also no longer necessary to add a catch-all `_ -> assert false` at the end.
- Functions like `run` are safer because the result type is checked.

## Drawbacks

Defining a GADT in the specification adds its own boilerplate: most functions now need explicit type annotations with universal quantifiers; the user must also define an existential type `packed_cmd` and wrap the command generators in it. Such STM tests are harder to write when one is not familiar with qcheck-stm.

And of course, another big drawback is that it changes the signature of STM specifications.

For these reasons, I open this PR as a draft, mostly to share the idea.

### DIFF
--- a/lib/STM.ml
+++ b/lib/STM.ml
@@ -57,16 +57,17 @@ let seq spec =
 
 
 
-type res =
-    Res : 'a ty_show * 'a -> res
+type 'a res = 'a ty_show * 'a
 
-let show_res (Res ((_,show), v)) = show v
+let show_res ((_,show), v) = show v
 
 (** The specification of a state machine. *)
 module type Spec =
 sig
-  type cmd
+  type 'a cmd
   (** The type of commands *)
+
+  type packed_cmd = Pack_cmd : 'a cmd -> packed_cmd
 
   type state
   (** The type of the model's state *)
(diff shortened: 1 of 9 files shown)


OlivierNicole:
> At the same time, as an end-user I probably don't want to care for such a `cmd` distinction... 🤷
> QCheck's `Test.make` is an example of a hook that sweeps an existential type parameter under the rug for 95% of end-users. I would love if we could pull off something similar here.

I would also love to sweep the existential under the rug, but this seems to me to clash with the fact that the spec writer needs to write a command generator. Since, as you say, the typechecker needs to unify all possible command types there.

Perhaps we could hide it in a function similar to `Test.make`, but what would the type of that function be?

OlivierNicole:
In the latest commit, the existential type is hidden behind a module with a nicer interface. The new compromise is that the sure must instantiate a small functor—see the new `src/atomic/stm_tests.ml`—and the distinction between `'r cmd` and `Cmd.any` still manifests itself when writing the generator.

The result is maybe syntactically closer to the existing, and hopefully the user does not have to deal with the cryptic error messages that GADTs can give.

OlivierNicole:
A good news is that the same tricks (not implemented in this PR, but I have it on the side) allow to get rid of the calls to `Obj.magic` in `Lin`:
https://github.com/ocaml-multicore/multicoretests/blob/c9930453f1ab8ee38075c48182fe070dc454e05d/lib/lin.ml#L422-L427
without changing the tests that use the high-level `Lin` interface in any way. Tests that use `Lin.Internal` do have to be adapted (in much the same way as in this PR), but given the unstable nature of that interface, I feel this will be less of an issue.

PR #476, opened 2024-09-09
Experiment: Add MSVC 32bit workflows

jmid:
MSVC 32bit workflows came up in #473 and may help catch 64bit-isms early.


jmid:
Closing this experimental PR for now.

Issue #466, opened 2024-07-05
Unexpected STM Sys test parallel test counterexample on s390x


jmid:
The `STM Sys test parallel test` on s390x with OCaml 5.1.0 found an unexpected counterexample:
https://ocaml-multicoretests.ci.dev:8100/job/2024-07-05/135811-ci-ocluster-build-a950ab
```
random seed: 462089430
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s STM Sys test sequential
[ ]    0    0    0    0 / 1000     0.0s STM Sys test sequential (generating)
[✓] 1000    0    0 1000 / 1000     1.5s STM Sys test sequential

[ ]    0    0    0    0 /  200     0.0s STM Sys test parallel
[✗]   58    0    1   57 /  200    10.3s STM Sys test parallel

[ ]    0    0    0    0 / 1000     0.0s STM Sys stress test parallel
[ ]  750    0    0  750 / 1000    48.5s STM Sys stress test parallel
[✓] 1000    0    0 1000 / 1000    63.1s STM Sys stress test parallel

--- Failure --------------------------------------------------------------------

Test STM Sys test parallel failed (4 shrink steps):

                                                              |                         
                                                         Readdir []                     
                                                 File_exists ["ccc"; "ddd"]             
                                     Mkfile (["ccc"; "ccc"; "ccc"; "bbb";... (truncated)
                                                      Mkdir ([], "ddd")                 
                                                     Mkfile ([], "aaa")                 
                                     Mkfile (["ccc"; "bbb"; "eee"; "eee";... (truncated)
                                                     File_exists ["ddd"]                
                                                      Mkdir ([], "eee")                 
                                                     Mkfile ([], "ddd")                 
                                                              |                         
                                   .----------------------------------------------------.
                                   |                                                    |                         
                              Readdir []                              Rmdir (["ccc"; "ccc"; "aaa"], "ddd")        
                            Readdir ["eee"]                                     Mkdir ([], "aaa")                 
                            Readdir ["bbb"]                                     Rmdir ([], "eee")                 
                           Rmdir ([], "eee")                   Mkfile (["bbb"; "aaa"; "ccc"; "bbb";... (truncated)
                           Mkdir ([], "aaa")                                    Mkdir ([], "bbb")                 
                           Rmdir ([], "ddd")                                    Rmdir ([], "ddd")                 
[...]

PR #478, opened 2024-09-11
Small cleanups to ease upstreaming

shym:
This PR brings a couple of small changes. In particular the main two commits are intended to make it a bit easier to upstream a workflow running the testsuite to the compiler:

- 665555c3 Define the `testsuite` and `internaltests` aliases
- 80ccd65f Silence all skipped tests

This will make it easier to run only the testsuite per se upstream, and to avoid the noise of the `Skipping ...` messages.

The other 3 commits are to use the new aliases in our CI, to let the `debug mode` message appear in debug runs and to remove an obsolete comment:

- 6662af3d Use the `testsuite` and `internaltests` aliases in CI
- cd5b374a Drop the v=0 parameter in CI debug runs
- fa43e243 Remove an obsolete comment

### DIFF
--- a/dune
+++ b/dune
@@ -16,13 +16,27 @@
  (package multicoretests)
  (deps (alias src/default)))
 
+; The main test alias
+(alias
+ (name testsuite)
+ (package multicoretests)
+ (deps
+  (alias_rec src/runtest)))
+
+; The internal tests alias
+(alias
+ (name internaltests)
+ (package multicoretests)
+ (deps
+  (alias_rec test/runtest)))
+
 ; The aliases to control what is run in CI
 ; It can either be the full test suite, or focus on a single test
 (alias
  (name ci)
  (package multicoretests)
  (deps
-  (alias_rec %{env:DUNE_CI_ALIAS=runtest})))
+  (alias_rec %{env:DUNE_CI_ALIAS=testsuite})))
   ; (alias_rec focusedtest)))
 
 ; @focusedtest
--- a/dune
+++ b/dune
@@ -18,8 +18,6 @@
 
 ; The aliases to control what is run in CI
 ; It can either be the full test suite, or focus on a single test
-; Aliases ci1 and ci2 are useful to split the test suite into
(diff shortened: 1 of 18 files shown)


dune
```diff
@@ -16,13 +16,27 @@
...
+  (alias_rec test/runtest)))
+
 ; The aliases to control what is run in CI
 ; It can either be the full test suite, or focus on a single test
 (alias
  (name ci)
  (package multicoretests)
  (deps
-  (alias_rec %{env:DUNE_CI_ALIAS=runtest})))
+  (alias_rec %{env:DUNE_CI_ALIAS=testsuite})))
```
jmid, line 39: Question out of curiosity: does this mean that one can avoid commenting in `; (alias_rec focusedtest)))` below and instead just set/override `dune_ci_alias` to `focusedtest` in `common.yml`? 
This is more a question to aid my understanding... :sweat_smile: 
It's probably preferable to keep the `dune`-file adjustments workflow in such a situation, as one has to adjust in the file anyway to point focus to a particular test... :shrug:

jmid (approved):
Small and sensible changes - LGTM!

jmid:
CI summary for cd5b374
- linux-ppc64le-5.2 found an unexpected STM Sys parallel test counterexample #466 

Out of 45 workflows, 1 test failed with a false alarm

dune
shym: That was indeed the idea of defining a `DUNE_CI_ALIAS` variable. But I’m not sure we’ve ever used that possibility, in particular since we have no way to set this variable to run the focused tests in multicoretests-ci. Maybe it should just be dropped (as part of this cleanup?)?

jmid: Thanks! I suggest we just keep the variable for now.

jmid:
Merging!

jmid:
- 32bit 5.2  triggered a segfault in `STM Domain.DLS test sequential` #446

Out of 46 workflows 1 failed with a genuine error

Issue #444, opened 2024-03-22
[ocaml5-issue] On Cygwin `Out_channel.flush` can cause `Syserror` when used in parallel


jmid:
A Cygwin 5.2 run triggered when merging #443 to main found an unexpected counterexample to `STM Out_channel parallel`:
https://github.com/ocaml-multicore/multicoretests/actions/runs/8361686850/job/22890359851
```
random seed: 292949458
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s STM Out_channel test sequential
[ ]    0    0    0    0 / 1000     0.0s STM Out_channel test sequential (generating)
[✓] 1000    0    0 1000 / 1000     2.9s STM Out_channel test sequential

[ ]    0    0    0    0 / 1000     0.0s STM Out_channel test parallel
[ ]  620    0    0  542 / 1000    57.1s STM Out_channel test parallel
[ ] 1116    0    0  966 / 1000   117.5s STM Out_channel test parallel (shrinking:    1.0011)
[✗] 1117    0    1  966 / 1000   129.8s STM Out_channel test parallel

--- Failure --------------------------------------------------------------------

Test STM Out_channel test parallel failed (1 shrink steps):

                                                              |                         
                                     Output_substring ("\225\185!\188Q\142\138", 0, 96) 
                                                       Output_byte 36                   
                                     Output_bytes "\158\199e\220\253\148\... (truncated)
                                                            Flush                       
                                                            Close                       
                                                   Output ("Z\197j", 3, 3)              
                                                          Open_text                     
                                                         Is_buffered                    
                                                    Set_binary_mode false               
                                              Output_substring ("{8)g", 18, 1)          
                                                           Length                       
                                                         Close_noerr                    
                                                          Open_text                     
                                                      Set_buffered true                 
                                                              |                         
                                   .----------------------------------------------------.
                                   |                                                    |                         
                              Close_noerr                                     Set_binary_mode true                
                               Open_text                                               Pos                        
                Output_substring ("\to\128N\186", 7, 6)                       Set_binary_mode true                
                            Output_char '9'                                      Output_byte 54                   
                         Set_binary_mode true                                        Length                       
                         Output_bytes "\182U"                                         Flush                       
                            Output_char '5'                    Output_substring ("\014\236\187\146^... (truncated)
              Output ("(|$\217\203\206\162t\235", 18, 4)                      Set_binary_mode true                
                                  Pos                                              Is_buffered                    
[...]

PR #483, opened 2024-11-01
Refine Out_channel string/bytes generator

jmid:
As part of carving a reproducer for #444, I realized that its string and bytes generators could be refined to have a bigger chance of producing strings stressing the channel boundary logic.

This quick little PR attempts such a refinement.
It is based on a `size_gen` generator that produces numbers around powers of 2:
```ocaml
let size_gen =
      Gen.(map2 (fun pos size_adj ->
                   let res = 1 lsl (pos-1) in
                   if size_adj < res then res-size_adj else res) (int_bound 18) (int_bound 10));;
val size_gen : int QCheck.Gen.t = <fun>
Gen.generate ~n:10 size_gen;;
- : int list = [65531; 16378; 65535; 128; 8185; 0; 249; 4; 59; 28]
Gen.generate ~n:10 size_gen;;
- : int list = [65531; 126; 1; 512; 131066; 13; 252; 4087; 16380; 120]
Gen.generate ~n:10 size_gen;;
- : int list = [0; 65533; 15; 8190; 4089; 4094; 8; 65528; 2044; 65530]
```


jmid:
This times out on MSVC bytecode and MinGW bytecode trunk, where I suspect the underlying buffer size to differ.

PR #491, opened 2024-12-13
Require qcheck-core.0.23

jmid:
This PR bumps the qcheck-core lower bound to 0.23.

This freshly-released version means the `Util.Pp.pp_fun_` output will contain properly escaped and quoted chars and strings, and thus more comprehensible - see, e.g., the expect test diff in ef6c0067.

As an added benefit, the new lower-bound means the Lin and STM show functions can be simplified,
as they no longer need to work around QCheck.Print's lack of quoting and escaping.

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -87,7 +87,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: c-cube/qcheck
-          ref: v0.22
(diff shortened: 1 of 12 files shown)


jmid:
CI summary for 2684c1b: all 31 workflows pass.

CI summary for merge to `main`: all 32 workflows pass.

PR #492, opened 2024-12-14
Use exponential

jmid:
This is an oversight forgotten in #491: 
`qcheck-core.0.23` includes an upstreamed `Gen.exponential` combinator (from STM even) - so we should then use it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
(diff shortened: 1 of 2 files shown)


jmid:
All 31 workflows green - merging

jmid:
The merge to `main` triggered one error:
- Linux trunk debug triggered an assertion error in src/weak/lin_tests_hashset.ml #493

Out of 32 workflows, 31 passed and 1 triggered a genuine error

PR #509, opened 2025-01-12
Change/Fix to use a symmetric barrier to synchronize domains

polytypic:
Unless I've misunderstood something, the idea is to synchronize the two domains before the two sequences of operations are run in parallel.  The previous implementation used an asymmetric approach where one domain only spins on a flag and the other only sets the flag.  That is not really guaranteed to synchronize the domains &mdash; the domain setting the flag might start running the operations before the domain spinning on the flag gets a chance to run.  This PR changes the code to use a symmetric barrier: a counter is initialized with 2, both domains decrement the counter, and then spin until the counter becomes zero.  This should improve the synchronization between the two domains.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## NEXT RELEASE
 
-- ...
+- #509: Change/Fix to use a symmetric barrier to synchronize domains
 
 ## 0.6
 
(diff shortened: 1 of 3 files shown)


jmid:
Thanks! Having a symmetric condition makes sense. I just ran some statistics to confirm that this does not influence the error finding behaviour (I wasn't expecting it to - but just to make sure).

Can I ask you to:
- make the corresponding change to `run_parallel` in `lib/lin_domain.ml` and
- add a CHANGES.md entry?

Thanks again! :pray:

jmid:
CI summary for 6c86871: all 34 workflows passed.
Merging...

jmid:
- `gh-pages` failed due to still using `setup-ocaml@v2` - fixed in #513 

Out of 35 workflows - 1 failed with a CI related error

Issue #510, opened 2025-01-13
Support testing effects based multicore primitives with QCheck STM


polytypic:
Currently QCheck STM spawns domains (`STM_domain`) or creates threads (`STM_thread`) on its own.

The problem with this is that it makes it difficult to test multicore primitives that potentially block using effects.

Here is an example Spec for testing `Rwlock`, a read-write lock implementation that uses effects:

```ocaml
module Spec = struct
  type cmd = Push of int | Pop_opt | Top_opt | Length

  let show_cmd = function
    | Push x -> "Push " ^ string_of_int x
    | Pop_opt -> "Pop_opt"
    | Top_opt -> "Top_opt"
    | Length -> "Length"

  module State = struct
    type t = int list

    let push x xs = x :: xs
    let pop_opt = function _ :: xs -> xs | [] -> []
    let top_opt = function x :: _ -> Some x | [] -> None
    let length = List.length
  end

  type state = State.t
  type sut = { stack : int Stack.t; rwlock : Rwlock.t }

  let arb_cmd _s =
    [
      Gen.int_range 1 1000 |> Gen.map (fun x -> Push x);
      Gen.return Pop_opt;
      Gen.return Top_opt;
      Gen.return Length;
    ]
    |> Gen.oneof |> make ~print:show_cmd

  let init_state = []
  let init_sut () = { stack = Stack.create (); rwlock = Rwlock.create () }
  let cleanup _ = ()

  let next_state c s =
    match c with
    | Push x -> State.push x s
    | Pop_opt -> State.pop_opt s
    | Top_opt -> s
    | Length -> s

  let precond _ _ = true

  let run c d =
    match c with
    | Push x ->
        Rwlock.wrlock d.rwlock;
        Stack.push x d.stack;
        Rwlock.unlock d.rwlock;
        Res (unit, ())
    | Pop_opt ->
        Rwlock.wrlock d.rwlock;
        let result = Stack.pop_opt d.stack in
        Rwlock.unlock d.rwlock;
        Res (option int, result)
    | Top_opt ->
        Rwlock.rdlock d.rwlock;
        let result = Stack.top_opt d.stack in
        Rwlock.unlock d.rwlock;
        Res (option int, result)
    | Length ->
        Rwlock.rdlock d.rwlock;
        let result = Stack.length d.stack in
        Rwlock.unlock d.rwlock;
        Res (int, result)

  let postcond c (s : state) res =
    match (c, res) with
[...]

PR #511, opened 2025-01-14
Introduce extended specs to allow wrapping command sequences

polytypic:
See #510.

Contrary to what #510 concluded with, I decided to introduce
- an extended `Spec` signature, `SpecExt`,
- a module of default specs, `SpecDefaults`,
- and extended functors, `STM_sequential.MakeExt`, `STM_domain.MakeExt`, and `STM_thread.MakeExt`.

The reason for this is that introducing `wrapped` variants of the various functions produced by the functors would have roughly meant duplicating all of the functions.  This approach is much more incremental and also potentially allows the `Spec` interface to the extended in the future.

The idea is that one always uses the `SpecDefaults`:

```ocaml
module MySpec = struct
  include SpecDefaults
  ...
end
```

This way new optional specs can be added as long as `SpecDetaults` can provide default implementations of such additions.

There is [a PR to Picos that uses this PR](https://github.com/ocaml-multicore/picos/pull/341).

TODO:
- [x] Agree this is the way to go
- [x] Document the API additions
- [x] Finish the implementation, including creating `STM_thread.MakeExt`, and making sure operations are wrapped in all cases
- [x] Update changelog

### DIFF
--- a/lib/STM.mli
+++ b/lib/STM.mli
@@ -164,7 +164,7 @@ sig
       {!SpecDefaults}. *)
 
   val cleanup : 'sut -> unit
-  (** [cleanup sut] just return [()]. *)
+  (** [cleanup sut] just returns [()]. *)
 
   val precond : 'cmd -> 'state -> bool
   (** [precond cmd state] just returns [true]. *)
--- a/lib/STM.mli
+++ b/lib/STM.mli
@@ -75,7 +75,9 @@ type res =
 val show_res : res -> string
 
 
-(** The specification of a state machine. *)
+(** The specification of a state machine.
+
+    See also {!SpecExt} and {!SpecDefaults}. *)
 module type Spec =
 sig
   type cmd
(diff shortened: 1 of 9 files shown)


jmid:
Thanks for this! :pray: 

I've played a bit with it locally, rewriting existing `STM` tests to use it. This made me realize that both `MakeExt` and `Make` clients can use `include SpecDefaults` which is a net-win IMO - even for users just sticking to `Make`. 
To ensure against regressions, I've also checked that this shouldn't cause error finding regressions, by rerunning the statistical tests.

I spotted locally that this would emit a compiler warning, but I believe you've fixed that in a recent (force) push.

Naming-wise I think
- `SpecExt` and `MakeExt` are reasonable for extended specs
- `SpecDefaults` is also fair - and also works well with the `STM.` quantification

My only reservation is about the `wrap` name. I don't think it describes well what is being wrapped.
Something like `wrap_cmd_seq` or `wrap_cmds` would be a bit more saying. I'm open to other suggestions - and not particularly eager about any of these two. The doc-string for (the binding-formerly-known-as) `wrap` should probably also describe a bit more what is being wrapped.

Otherwise, I'm OK with this :slightly_smiling_face: :+1:

polytypic:
I decided to rename to `wrap_cmd_seq`, because `wrap_cmd` might give the impression of wrapping each and ever command separately.

lib/STM.mli
```diff
@@ -126,6 +128,50 @@ sig
...
+      This signature may be extended in the future with new specifications that
+      can be given defaults via {!SpecDefaults}. *)
+
+  include Spec
+
+  val wrap_cmd_seq : (unit -> 'a) -> 'a
+  (** [wrap_cmd_seq thunk] must call [thunk ()] and return or raise whatever
+      [thunk ()] returned or raised.  [wrap_cmd_seq] is used to wrap the
+      execution of the generated command sequences.  [wrap_cmd_seq] is useful,
+      e.g., to handle effects performed by blocking primitives. *)
```
jmid, line 144:
```suggestion
  (** [wrap_cmd_seq] is used to wrap the execution of the generated command
      sequences. [wrap_cmd_seq] is useful, e.g., to handle effects performed by
      blocking primitives. [wrap_cmd_seq thunk] must call [thunk ()] and return
      or raise whatever [thunk ()] returned or raised. *)
```
Nit: Moving the first sentence to the end improves the information flow I think:
Offering a general description first and finishing with requirement details.

lib/STM.mli
```diff
@@ -126,6 +128,50 @@ sig
...
+          (* ... *)
+        end
+      ]}
+
+      This way the spec module can usually just continue working after new
+      specifications have been added to {!SpecExt} with defaults in
+      {!SpecDefaults}. *)
+
+  val cleanup : 'sut -> unit
+  (** [cleanup sut] just return [()]. *)
```
jmid, line 167:
```suggestion
  (** [cleanup sut] just returns [()]. *)
```
(aligning it with the `precond` doc-string)

jmid (approved):
LGTM. I spotted a couple of minor documentation nits while proof-reading.

jmid:
CI summary for 263ae96: all 34 workflows passed.
Merging...

PR #512, opened 2025-01-14
Use setup-ocaml 3 for opam workflow

jmid:
We started observing opam workflow failures, e.g., on #509, due to `ubuntu-latest` switching to Ubuntu 24.04.
Switching to setup-ocaml@v3 fixes this. See https://github.com/ocaml/setup-ocaml/issues/872


jmid:
- the github pages workflow failed as it is also using setup-ocaml@v2

Out of 35 workflows, 1 failed with a non-test-suite failure

PR #469, opened 2024-08-27
Add gc tests

jmid:
This PR adds initial tests of the `Gc` module - and in particular testing of `Gc.compact`.

Exercising `Gc.counters` already triggered a known issue locally, which has recently been fixed on `trunk`: https://github.com/ocaml/ocaml/pull/13370 I therefore expect the parallel test to fail on 5.2.0 and earlier.

This can be extended (and improved) in many ways adding (big)arrays, Weak, Ephemeron, custom blocks, finalizers, and dynamic Gc_ctrl changes.

PR #513, opened 2025-01-14
CI updates

jmid:
This PR lumps 3 CI updates into one:
- The first 2 commits removes the 5.0 workflows (and badges). After the merge of the Gc tests #469, 9/9 of them failed the scheduled run, creating lots of notification email noise. Rather than spending time adjusting them to accept misbehavior on 5.0, now seems a natural point to stop running them and save some CPU cycles in the process.
- The next 2 commits removes the 5.1 musl workflow (and badge) which is failing compilation with `zstd.h not found`. This is a known bug and the opam 5.1.1 opam package was explicitly back-patched in https://github.com/ocaml/opam-repository/pull/25828 to handle this error. In choosing between continuing to fail, spending time on patching in our opam-free setup, and removing the musl-workflow from 5.1, the latter seems most sane.
- The last commit updates the gh-pages workflow to also use `setup-ocaml@v3` - an oversight from #512


jmid:
CI summary for 91072aa
- Cygwin trunk timed out after 7529.2s shrinking in `Lin Dynarray test with Domain` #514 

Out of 34 workflows, 1 failed with a false alarm.

jmid:
CI summary for merge to `main`: all 35 workflows passed!

PR #516, opened 2025-01-15
Lin bytes adjustments

### LINKED ISSUE #515: Long runtime in Lin Bytes test with Thread on Cygwin
While studying the CI log in connection with #514 I spotted another test taking a long time on Cygwin with trunk:
https://github.com/ocaml-multicore/multicoretests/actions/runs/12777231067/job/35617614765?pr=513
```
[ ]    0    0    0    0 / 1000     0.0s Lin Bytes test with Thread
[ ]    5    0    0    5 / 1000  3043.2s Lin Bytes test with Thread
[ ]  625    0    0  625 / 1000  3103.3s Lin Bytes test with Thread
[✓] 1000    0    0 1000 / 1000  3139.5s Lin Bytes test with Thread
```

jmid:
Looking at the `Lin Bytes` test I noticed an embarrasing oversight: it was using the uniform `int` generators.
- The first commit thus adjusts them to use small ones - meaning we start triggering counterexamples - and thus turning the test into a negative one. Bytes is unsafe to use in parallel after all :shrug: 
- The second commit reduces the `Thread` count to 250, like, e.g., the similar `Stack Thread` test
- The third commit adds `Bytes.set` - another unexpected oversight :grimacing: 

(In fairness the STM Bytes tests are free from both of the above oversights)

Fixes #515

### DIFF
--- a/src/bytes/lin_tests.ml
+++ b/src/bytes/lin_tests.ml
@@ -7,7 +7,7 @@ module BConf = struct
   let cleanup _ = ()
 
   open Lin
-
+  let int,string = nat_small, string_small_printable
   let api = [
     val_ "Bytes.get"         Bytes.get         (t @-> int @-> returning_or_exc char);
     val_ "Bytes.sub_string"  Bytes.sub_string  (t @-> int @-> int @-> returning_or_exc string);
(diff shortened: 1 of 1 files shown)


jmid:
CI summary: all 34 workflows passed.
Merging...

jmid:
CI summary for merge to `main`
- MSVC bytecode 5.3 failed to trigger a Lin Bytes counterexample #518 

Out of 35 workflows, 1 failed with a false alarm

Issue #528, opened 2025-01-19
[ocaml5-issue] `dummy found!` in `Lin Dynarray stress test with Domain` on musl trunk


jmid:
On #521 the musl trunk workflow triggered a `dummy found!` error exiting the `Lin Dynarray` stress test:
https://github.com/ocaml-multicore/multicoretests/actions/runs/12846368030/job/35821476446?pr=521
```
random seed: 225104823
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s Lin Dynarray test with Domain
[ ]    0    0    0    0 / 1000     0.0s Lin Dynarray test with Domain (generating)
[✓]    2    0    1    1 / 1000     2.1s Lin Dynarray test with Domain

dummy found!
File "_build/.dune/default/src/dynarray/dune", line 3, characters 7-16:
3 |  (name lin_tests)
           ^^^^^^^^^
(cd _build/default/src/dynarray && ./lin_tests.exe --verbose)
Command exited with code 1.
[ ]    0    0    0    0 / 1000     0.0s Lin Dynarray stress test with Domain
```

This is caused by these lines implementing a `get` instrumented with a tag check:
https://github.com/ocaml-multicore/multicoretests/blob/a166b242ad1f44c3ff37f07bae36ef35460f0290/src/dynarray/lin_tests.ml#L11-L13
IIUC, this means the test is observing a non-int dummy value from the unboxed implementation from https://github.com/ocaml/ocaml/pull/12885.
A code comment in stdlib/dynarray.ml reads:
```
   [...] This dummy is *not* type-safe, it is not a valid value of type ['a], so
   we must be very careful never to return it to the user. After
   accessing any element of the array, we must check that it is not
   the dummy. In particular, this dummy must be distinct from any
   other value the user could provide -- we ensure this by using
   a dynamically-allocated mutable reference as our dummy.
```

At the same time, the module is documented to be unsafe for parallel usage:
```
Unsynchronized accesses

    Unsynchronized accesses to dynamic arrays are a programming error.

   Concurrent accesses to dynamic arrays must be synchronized
   (for instance with a {!Mutex.t}). Unsynchronized accesses to
   a dynamic array are a programming error that may lead to an invalid
   dynamic array state, on which some operations would fail with an
   [Invalid_argument] exception.
```

Can we conclude that parallel usage may be type unsafe? Ping @OlivierNicole: WDYT? 🤔

OlivierNicole:
[...]

PR #529, opened 2025-01-27
Improve reporting in Dynarray lin test

OlivierNicole:
By raising an exception rather than aborting the program. This should help with #528 when the bug surfaces again.

### DIFF
--- a/src/dynarray/lin_tests.ml
+++ b/src/dynarray/lin_tests.ml
@@ -10,7 +10,7 @@ module Dynarray_api = struct
 
   let get_check a i =
     let v = Dynarray.get a i in
-    if not (Obj.is_int (Obj.repr v)) then (Printf.eprintf "dummy found!\n%!"; exit 1) else v
+    if not (Obj.is_int (Obj.repr v)) then failwith "dummy found!" else v
 
   let api =
     (*let int_not_too_big = int_bound 2048 in*)


jmid:
Thanks! Unfortunately I don't think this will make a big difference in helping uncover the bug.
To illustrate, I've patched the `Lin Dynarray` test specification to always fail:
```diff
diff --git a/src/dynarray/lin_tests.ml b/src/dynarray/lin_tests.ml
index 33680a0b..54982172 100644
--- a/src/dynarray/lin_tests.ml
+++ b/src/dynarray/lin_tests.ml
@@ -10,7 +10,7 @@ module Dynarray_api = struct
 
   let get_check a i =
     let v = Dynarray.get a i in
-    if not (Obj.is_int (Obj.repr v)) then (Printf.eprintf "dummy found!\n%!"; exit 1) else v
+    if (*not (Obj.is_int (Obj.repr v))*) true then failwith "dummy found!" else v
 
   let api =
     (*let int_not_too_big = int_bound 2048 in*)
```

With this patch, the test always fails on the first `Lin` input generated (they all contain a `get` I guess), however after shrinking I have yet to see one with `get` in its reported counterexample:
```
$ dune exec src/dynarray/lin_tests.exe -- -v
random seed: 497616361
[✓]    1    0    1    0 / 1000     1.5s Lin Dynarray test with Domain
[✓] 1000    0    0 1000 / 1000     3.1s Lin Dynarray stress test with Domain

Negative test Lin Dynarray test with Domain failed as expected (27 shrink steps):

                              |          
                              |          
                   .---------------------.
                   |                     |          
                reset t         append_seq t <0; 0> 


Messages for test Lin Dynarray test with Domain:

  Results incompatible with sequential execution

                                                                 |                              
                                                                 |                              
                                  .-------------------------------------------------------------.
                                  |                                                             |                              
                          reset t : Ok (())                        append_seq t <0; 0> : Error (Invalid_argument("Array.blit"))

success (ran 2 tests)
```

- for the negative parallel Lin test, `get` may indeed raise an exception (Lin doesn't let us express which ones are OK and which are not) and
- for the parallel Lin stress test, `get` is therefore only raising an exception (as expected) and thus passing that test.

Overall, as a result this could potentially hide occurrences of the bug!

As you've earlier suggested on #528, I think the best option is to try reproducing with an STM test that allows to specify acceptable behaviour more specifically... :thinking:

OlivierNicole:
Right, I see how that couldn’t work! I have botched up the STM test to make it closer to the Lin test that uncovered the bug here: https://github.com/OlivierNicole/multicoretests/tree/track-dynarray-bug and am now trying to reproduce the bug.

OlivierNicole:
Closing as there is no clear change that can be included into the main branch at this point.

PR #531, opened 2025-02-17
CI: Update `dune` to support `trunk`

shym:
Unfortunately, this is not part of a released version yet.

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -128,7 +128,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: ocaml/dune
-          ref: 3.16.0
+          ref: 2de53980efa731eacf5475139d9a5a218f9752a0
           path: dune
         if: steps.cache.outputs.cache-hit != 'true'
 


jmid (approved):
LGTM, thanks! :slightly_smiling_face: :+1: 

The `trunk` opam installation workflow is still failing, but rather than introducing a workaround for that, I think we should just wait for a dune release with the patch.

I guess the `run-multicoretests` workflow on ocaml/ocaml will also fail for the same reason. Since it is not used much, unless someone gets bitten by it, we should consider postponing fixing that until a dune release with the patch. Otherwise we will end up with 2 ocaml/ocaml PRs (one for temporarily specifying a git hash like here + a later one to set the released version), creating noise for little benefit. :shrug:

shym:
> LGTM, thanks! 🙂 👍

Thanks :pray:

> The `trunk` opam installation workflow is still failing, but rather than introducing a workaround for that, I think we should just wait for a dune release with the patch.

We could consider disabling temporarily the Opam workflow on trunk, I suppose.

> I guess the `run-multicoretests` workflow on ocaml/ocaml will also fail for the same reason. Since it is not used much, unless someone gets bitten by it, we should consider postponing fixing that until a dune release with the patch. Otherwise we will end up with 2 ocaml/ocaml PRs (one for temporarily specifying a git hash like here + a later one to set the released version), creating noise for little benefit. 🤷

I agree!

jmid:
CI summary for merge to `main`
- opam installation on 5.4.0/trunk failed, causing 7 other opam installation workflows to cancel

Out of 35 workflows, 1 failed with a genuine error and 7 others were cancelled

Issue #526, opened 2025-01-17
Deadlock in `Lin Bytes test with Thread` on Cygwin


jmid:
Opening this to track it separately from #520, as it will allow to distinguish occurrences of one from the other.
Also, in comparison to #520, this  on a `Thread`s test - and not involving shrinking.

> On the merge to main of #519 for the Cygwin 5.3 workflow we just saw excessive time spent in Lin Bytes test with Thread
> https://github.com/ocaml-multicore/multicoretests/actions/runs/12808801661/job/35712309410
>
> ```
> random seed: 366364260
> generated error fail pass / total     time test name
>
> [ ]    0    0    0    0 / 2500     0.0s Lin Bytes test with Domain
> [ ]    0    0    0    0 / 2500     0.0s Lin Bytes test with Domain (generating)
> [ ]   37    0    0   37 / 2500    60.2s Lin Bytes test with Domain (shrinking:   15.0003)
> [ ]   37    0    0   37 / 2500   122.0s Lin Bytes test with Domain (shrinking:   18.0006)
> [ ]   37    0    0   37 / 2500   182.5s Lin Bytes test with Domain (shrinking:   20.0008)
> [ ]   37    0    0   37 / 2500   245.1s Lin Bytes test with Domain (shrinking:   23.0005)
> [ ]   37    0    0   37 / 2500   308.1s Lin Bytes test with Domain (shrinking:   25.0002)
> [ ]   37    0    0   37 / 2500   371.9s Lin Bytes test with Domain (shrinking:   28.0010)
> [ ]   37    0    0   37 / 2500   433.4s Lin Bytes test with Domain (shrinking:   30.0013)
> [ ]   37    0    0   37 / 2500   493.5s Lin Bytes test with Domain (shrinking:   32.0012)
> [ ]   37    0    0   37 / 2500   556.0s Lin Bytes test with Domain (shrinking:   34.0002)
> [ ]   37    0    0   37 / 2500   617.4s Lin Bytes test with Domain (shrinking:   35.0012)
> [ ]   37    0    0   37 / 2500   677.9s Lin Bytes test with Domain (shrinking:   36.0015)
> [ ]   37    0    0   37 / 2500   741.5s Lin Bytes test with Domain (shrinking:   37.0013)
> [ ]   37    0    0   37 / 2500   806.4s Lin Bytes test with Domain (shrinking:   39.0004)
[...]

PR #540, opened 2025-03-21
Improve context switching chance in `Lin_thread` and `STM_thread` modes

### LINKED ISSUE #338: Improve reproducibility of Lin/STM Thread modes
The Lin and STM Thread modes do not work as well as the `Domain` modes.
Failures to trigger an issue (or unexpectedly triggering one for a change) adds noise to our CI results.

Some past history:
- The Lin Thread mode was added in #1
- #24 introduced the idea of using a statistical test to guide improvements
- #88 adds a corresponding Thread mode for STM 
 #99 tried out some ideas for improvements
  - generating symbolic `Thread.yield`s
  - overriding the `Stdlib` with one that triggers more `Thread.yield`s than usual
 
As part of the latter, I spent some more time banging my head against the `Thread` issue.
In the end I used custom `runtime_events` to record when `Thread`s would `spawn` to understand what we were observing.
This meant I could get a nice diagram like this:

thread-screenshot (1)
Which confirmed to me that most of our `Thread` tests indeed had two `Threads` running concurrently (starting and ending roughly at the same time)

I then went one step further and added custom events for `run` of one `cmd` (a small and fast function). This is the result:

run_cmd_screenshot
The little arrows mark simultaneous call and return of `run`. There are two next to each other, because `Thread.yield` triggers right after one (on Thread 0) and then the next happens (on Thread 1). This tells me that the `Thread`s are also interleaving - it just happens too deterministically - in between each `run` - but very very very rarely in the middle of it.

The above may be helpful to guide further improvement attempts.

jmid:
This fixes #338 - or at least gives it a good kick in the right direction.

Both `Lin_thread` and `STM_thread` are currently marked as experimental - and for good reason:
The chance of them triggering unfortunate, error-triggering `Thread` context switches is currently slim.
In hindsight, releasing them earlier - even as `experimental` - was probably premature.

Recall that a `Thread` context switch may happen
- at allocation points and
- at safepoints https://github.com/ocaml/ocaml/blob/trunk/asmcomp/polling.ml

IIUC: 
- The latter is the case for the native code backend.
- The bytecode backend instead uses a `CHECK_SIGNALS` instruction inserted in `while` and `for` loops.
- Both backends utilize a "tick thread" to context switch every 50ms to avoid starvation https://github.com/ocaml/ocaml/blob/trunk/otherlibs/systhreads/st_stubs.c

This PR improves on the situation by utilizing a terrific hack by @edwintorok [in the form of a `Gc.Memprof` callback](https://github.com/ocaml/ocaml/issues/13430):
Upon allocations, with a certain probability, `Memprof` will trigger a callback which can be utilized to trigger an explicit `Thread.yield`.

To get an idea how well this works, I've run stats on our usual {int,int64} x {ref,CList} negative tests, repeating each one 10.000 times, and recording how many of these failed the "interface is `Thread`-safe" property. Since the `Thread` failures are still relatively rare, each of the 10.000 test inputs is `Util.repeat`ed 25 times, to bring up the probability of a failure.

For `CList`, the `Domain`-unsafe `add_node` does not do any allocations in the central "unsafe window", effectively making it `Thread` safe. I've therefore introduced an even worse `add_node_thread` by moving an allocation to this part,
for the purpose of testing...

Each test is run on both 5.3.0, 5.2.0, and 4.14.2 - with only 5.2.0 not having `Gc.Memprof` support: It is available in 4.14 and was restored in 5.3. By running on 5.2 the `Memprof.{start,stop}` calls thereby degrade to essentially noops, and enable an easy comparison with/without the improvement.

#### STM numbers with `Util.repeat 25`:
```
                        5.3.0  5.3.0-bc  5.2.0  5.2.0-bc  4.14.2  4.14.2-bc

int ref STM Thread         0       0        0        0        0       0

int64 ref STM Thread     925       0(!)    34        0      899       0(!)

CList int STM.thread     260     260        0        0      275     268

CList int64 STM.thread   277     274        0        0      274     280
```

#### Lin numbers with `Util.repeat 25`:
```
                        5.3.0  5.3.0-bc  5.2.0  5.2.0-bc  4.14.2  4.14.2-bc

int ref Lin Thread         0       0        0        0        0       0

int64 ref Lin Thread     796       0(!)     1        0      804       0(!)

int CList Lin Thread     161     122        0        0      115     123

int64 CList Lin Thread   124     139        0        1      143     137
```

From the above it is clear that 5.3 and 4.14 works remarkably better than the current version.
Even in bytecode mode this represents a good step forward. However, for some reason, in bytecode mode
both `Lin_thread` and `STM_thread` fail to trigger an `int64 ref` issue. I've not yet figured out why this is the case.

Overview of commits
- The first two commits makes the above change to the two `Lin` and `STM` modes
- Because the fix works relatively well, the next commits removes the previous `experimental` alert attributes
- OCaml 5.0-5.1 raise an exception upon calling `Memprof.{start,stop}` so the 5th commit wraps them in a handler
- The next 2 commits improves documentation
- A bunch of commits adjust the `src/neg_tests` tests to adjust for the improved behaviour
- For anyone interested commits ad1d5e4a45d647a42c73fc09c57f7572afe39da2 and  a4a635cc8f5b32189a61b6a4dce04ebe20c46b72 contain the source for the above stats - I plan to remove these before merging.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,9 @@
 
 ## NEXT RELEASE
 
-- ...
+- #540: Significantly increase the probability of context switching in `Lin_thread`
+        and `STM_thread` by utilizing `Gc.Memprof` callbacks. Avoid on 5.0-5.2
+        without `Gc.Memprof` support.
 
 ## 0.7
 
--- a/lib/STM_thread.ml
+++ b/lib/STM_thread.ml
@@ -11,6 +11,14 @@ module MakeExt (Spec: SpecExt) = struct
 
   let arb_cmds_triple = arb_cmds_triple
 
+  let alloc_callback _src =
+    Thread.yield ();
+    None
+
+  let yield_tracker =
+    Gc.Memprof.{ null_tracker with alloc_minor = alloc_callback;
+                                   alloc_major = alloc_callback; }
+
   (* [interp_sut_res] specialized for [Threads] *)
   let rec interp_sut_res sut cs = match cs with
     | [] -> []
@@ -23,13 +31,15 @@ module MakeExt (Spec: SpecExt) = struct
   let agree_prop_conc (seq_pref,cmds1,cmds2) =
     let sut = Spec.init_sut () in
     let obs1,obs2 = ref (Error ThreadNotFinished), ref (Error ThreadNotFinished) in
+    let _ = Gc.Memprof.start ~sampling_rate:1e-3 ~callstack_size:0 yield_tracker in
     let pref_obs = Spec.wrap_cmd_seq @@ fun () -> interp_sut_res sut seq_pref in
     let wait = ref true in
     let th1 = Thread.create (fun () -> Spec.wrap_cmd_seq @@ fun () -> while !wait do Thread.yield () done; obs1 := try Ok (interp_sut_res sut cmds1) with exn -> Error exn) () in
     let th2 = Thread.create (fun () -> Spec.wrap_cmd_seq @@ fun () -> wait := false; obs2 := try Ok (interp_sut_res sut cmds2) with exn -> Error exn) () in
-    let ()   = Thread.join th1 in
-    let ()   = Thread.join th2 in
-    let ()   = Spec.cleanup sut in
+    Thread.join th1;
+    Thread.join th2;
+    Gc.Memprof.stop ();
+    Spec.cleanup sut;
     let obs1 = match !obs1 with Ok v -> v | Error exn -> raise exn in
     let obs2 = match !obs2 with Ok v -> v | Error exn -> raise exn in
     check_obs pref_obs obs1 obs2 Spec.init_state
--- a/lib/STM_thread.ml
+++ b/lib/STM_thread.ml
@@ -32,7 +32,7 @@ module MakeExt (Spec: SpecExt) = struct
     let sut = Spec.init_sut () in
     let obs1,obs2 = ref (Error ThreadNotFinished), ref (Error ThreadNotFinished) in
     (* Gc.Memprof.{start,stop} raises Failure on OCaml 5.0 and 5.1 *)
-    (try ignore (Gc.Memprof.start ~sampling_rate:1e-3 ~callstack_size:0 yield_tracker) with Failure _ -> ());
+    (try ignore (Gc.Memprof.start ~sampling_rate:1e-1 ~callstack_size:0 yield_tracker) with Failure _ -> ());
     let pref_obs = Spec.wrap_cmd_seq @@ fun () -> interp_sut_res sut seq_pref in
     let wait = ref true in
     let th1 = Thread.create (fun () -> Spec.wrap_cmd_seq @@ fun () -> while !wait do Thread.yield () done; obs1 := try Ok (interp_sut_res sut cmds1) with exn -> Error exn) () in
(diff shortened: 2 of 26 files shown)


jmid:
FTR, the PR initially used the magic constant of `1e-3` samples per word, after earlier experimenting with higher frequencies of interrupts. Especially in bytecode mode, those seemed to result in too big of a slowdown.
Feedback from @edwintorok however made me realize that I hadn't explored the trade off between the frequency of interrupts (more `Thread.yield`'s) and the `Util.repeat` count (raising the error rate by repetition). A second look at the code furthermore revealed that all but the stats were still using `Util.repeat 100` - a good catch.

I have experimented a bit more by timing `src/neg_tests/stm_tests_clist_thread_stats.exe` under different `sampling_rate`s  and `rep_count`s. Turning up either of these two increases the error rate (and hence bug finding ability) however doing so increases the test running time (because of more callbacks and more repeated executions, respectively), which will be noticeable for positive tests that run, e.g., 1000 iterations. To keep the experiment manageable I've measured on native code where the stat benchmark takes ~5min compared to ~30min under bytecode...

The initial `~sampling_rate:1e-3` `rep_count:25`
```
CList int STM.thread 313 / 10000
CList int64 STM.thread 294 / 10000
265.02user 38.18system 4:52.81elapsed 103%CPU (0avgtext+0avgdata 31680maxresident)k
```

Now, with `~sampling_rate:1e-1` `rep_count:3` the wall clock time is just ~7s slower out of ~5m but with a ~5x better error rate:
```
CList int STM.thread 1564 / 10000
CList int64 STM.thread 1613 / 10000
294.49user 7.35system 5:01.87elapsed 99%CPU (0avgtext+0avgdata 31880maxresident)k
```

The above seems like a good compromise between not affecting time usage significantly, while giving the error rate a significant bump upwards :slightly_smiling_face: 

Both of the above were run with OCaml 5.3.0. For comparison, I've timed the same run on OCaml 5.2.0 (where `Memprof` is not supported) with rep_count:100 to mimic the state of affairs before this PR:
```
CList int STM.thread 2 / 10000
CList int64 STM.thread 2 / 10000
412.03user 181.22system 8:57.95elapsed 110%CPU (0avgtext+0avgdata 31732maxresident)k
```
Going from ~9m to ~5m wall clock time shows the `Memprof` solution is not only better error-rate-wise but also much faster (the many repetitions is able to trigger issues in 2 out of 10000 test cases).

I also timed the run on OCaml 5.2.0 with rep_count:3 to assess the overhead of `Memprof`:
```
CList int STM.thread 0 / 10000
CList int64 STM.thread 0 / 10000
179.91user 5.55system 3:05.20elapsed 100%CPU (0avgtext+0avgdata 31676maxresident)k
```
so the `Memprof` callbacks add a ~2m (~5m - ~3m) ~66% overhead on this particular test. Non-allocating tests will naturally incur less of an overhead, while heavily allocating tests may incur a higher one.

Finally, for completeness, below follows the repeated stats from the initial PR description.
These confirm that the error rate further improves across the board - except on 5.2 (and 5.0+5.1),
due to the reduced repetition count and no Memprof support.

#### Revised STM numbers with `sampling_rate:1e-1` and `Util.repeat 3`
```
                        5.3.0  5.3.0-bc  5.2.0  5.2.0-bc  4.14.2  4.14.2-bc

int ref STM Thread         0       0        0        0        0       0

int64 ref STM Thread    3826       0        7        0     3891       0

CList int STM.thread    1561    1576        0        0     1639    1650

CList int64 STM.thread  1558    1572        0        0     1579    1662
```

#### Revised Lin numbers with `sampling_rate:1e-1` and `Util.repeat 3`
```
                        5.3.0  5.3.0-bc  5.2.0  5.2.0-bc  4.14.2  4.14.2-bc

int ref Lin Thread         0       0        0        0        0       0

int64 ref Lin Thread    3407       0        0        0     3395       0

int CList Lin Thread     702     710        0        0      700     725

int64 CList Lin Thread   721     665        0        0      677     686
```

jmid:
CI summary for ca10290: All 36 workflows passed.

CI summary for 292a3e0:
- Cygwin trunk timed out in `Lin Bytes test with Thread` #526 

Out of 36 workflows, 1 failed with a genuine (supposedly Cygwin) error

jmid:
CI summary for 0ef30bb: all 36 workflows passed. 
I'll merge.

jmid:

Issue #541, opened 2025-03-25
Test "Lin Bytes test with Domain" is flaky


glondu:
The test "Lin Bytes test with Domain" does not always behave the same, causing random failures in Debian:
- https://buildd.debian.org/status/fetch.php?pkg=ocaml-multicoretests&arch=s390x&ver=0.7-1%2Bb2&stamp=1742741819&raw=0
- https://people.debian.org/~sanvila/build-logs/202503/ocaml-multicoretests_0.7-1_amd64-20250317T133603.533Z

The failure seldom happens, and I couldn't reproduce it myself (even using the same nonce).

glondu:
Reported in Debian as https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1101243

jmid:
Thanks for sharing, acknowledged! 
If I am reading the logs correctly, you've now observed this on both `amd64` and `s390x`, right?

Are these run in some form of containerized environment?
Also are they running on native amd64/s390x hardware or via, e.g., qemu?

"Lin Bytes test with Domain" has been a pain point also in #520. The test was furthermore extended in #521. 
As more bindings are targeted, the chance of hitting a bad combination decreases. It could be bettered by introducing weights. Alternatively, most other tests have been rewritten using `STM`. An alternative path could thus be to port the #521 extensions to the existing `src/bytes/stm_tests.ml` test.

glondu:
> If I am reading the logs correctly, you've now observed this on both `amd64` and `s390x`, right?

Right.

> Are these run in some form of containerized environment?

Yes. I don't know exactly the setup for the amd64 occurrence, but the s390x one surely ultimately boils down to either chroot or unshare.

> Also are they running on native amd64/s390x hardware or via, e.g., qemu?

I don't know the exact setup for the amd64 occurrence, but the s390x one occured on this machine: https://db.debian.org/machines.cgi?host=zandonai (I would say "native", if that means anything for s390x).

sanvila:
Hello. I am the one who filed the original report in Debian.

While we are at it: I've also noticed that the tests fail 100% of the time on the single-CPU instances where I tried to build the package.

Am I right to think, given the package name (multicoretests) that it does not make sense to run the tests on single-CPU systems? (I guess this could be fixed easily in debian/rules).

Or maybe the tests themselves could detect that they are being executed on a single-cpu system and do nothing in such case?

[...]

Issue #520, opened 2025-01-16
Timeout due to excessive shrinking in `Lin Bytes test with Domain` on Cygwin


jmid:
On the merge to `main` of #517 we saw a timeout on Cygwin trunk due to excessive shrinking in the `Lin Bytes test with Domain` test:
https://github.com/ocaml-multicore/multicoretests/actions/runs/12796162913/job/35675203217
```
random seed: 306961601
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s Lin Bytes test with Domain
[ ]    0    0    0    0 / 1000     0.0s Lin Bytes test with Domain (generating)
[ ]   85    0    0   85 / 1000    62.4s Lin Bytes test with Domain (shrinking:    3.0014)
[ ]   85    0    0   85 / 1000   125.7s Lin Bytes test with Domain (shrinking:    5.0022)
[ ]   85    0    0   85 / 1000   186.3s Lin Bytes test with Domain (shrinking:    6.0025)
[ ]   85    0    0   85 / 1000   250.2s Lin Bytes test with Domain (shrinking:    7.0002)
[ ]   85    0    0   85 / 1000   312.7s Lin Bytes test with Domain (shrinking:    9.0003)
[ ]   85    0    0   85 / 1000   374.2s Lin Bytes test with Domain (shrinking:    9.0030)
[ ]   85    0    0   85 / 1000   434.2s Lin Bytes test with Domain (shrinking:   12.0004)
[ ]   85    0    0   85 / 1000   494.3s Lin Bytes test with Domain (shrinking:   19.0051)
[ ]   85    0    0   85 / 1000   554.5s Lin Bytes test with Domain (shrinking:   31.0010)
[ ]   85    0    0   85 / 1000   614.6s Lin Bytes test with Domain (shrinking:   41.0003)
[ ]   85    0    0   85 / 1000   678.1s Lin Bytes test with Domain (shrinking:   44)
[ ]   85    0    0   85 / 1000   780.8s Lin Bytes test with Domain (shrinking:   46)
[ ]   85    0    0   85 / 1000   881.0s Lin Bytes test with Domain (shrinking:   47)
[ ]   85    0    0   85 / 1000   969.1s Lin Bytes test with Domain (shrinking:   48)
[ ]   85    0    0   85 / 1000  1045.8s Lin Bytes test with Domain (shrinking:   49)
[...]

Issue #518, opened 2025-01-15
Lin Bytes negative test fails to trigger on MSVC bytecode


jmid:
#516 changed the Lin Bytes test, turning it into a negative test.

The merge to `main` of the PR then failed to find a counterexample within 1000 attempts on MSVC bytecode 5.3:
https://github.com/ocaml-multicore/multicoretests/actions/runs/12791598373/job/35660087124
```
random seed: 278988859
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s Lin Bytes test with Domain
[ ]    0    0    0    0 / 1000     0.0s Lin Bytes test with Domain (generating)
[ ]   75    0    0   75 / 1000    60.1s Lin Bytes test with Domain
[ ]  343    0    0  343 / 1000   121.0s Lin Bytes test with Domain
[ ]  926    0    0  926 / 1000   181.0s Lin Bytes test with Domain
[✗] 1000    0    0 1000 / 1000   188.4s Lin Bytes test with Domain

[ ]    0    0    0    0 /  250     0.0s Lin Bytes test with Thread
[ ]   69    0    0   69 /  250   243.1s Lin Bytes test with Thread
[✓]  250    0    0  250 /  250   300.4s Lin Bytes test with Thread

--- Failure --------------------------------------------------------------------

Test Lin Bytes test with Domain failed:

Negative test Lin Bytes test with Domain succeeded but was expected to fail
================================================================================
failure (1 tests failed, 0 tests errored, ran 2 tests)
File "src/bytes/dune", line 12, characters 7-16:
12 |  (name lin_tests)
            ^^^^^^^^^
(cd _build/default/src/bytes && ./lin_tests.exe --verbose)
Command exited with code 1.
```

We should just increase the `count` for a start, as it is not taking particularly long to execute - and doesn't cost anything for platforms finding a counterexample earlier :shrug:

jmid:
Reopening this as this happened on the push to `main` in connection with the 0.7 release for MSVC bytecode trunk:
https://github.com/ocaml-multicore/multicoretests/actions/runs/12891458000/job/35943440732
```
random seed: 318271471
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 5000     0.0s Lin Bytes test with Domain
[ ]    0    0    0    0 / 5000     0.0s Lin Bytes test with Domain (generating)
[ ]  609    0    0  609 / 5000    88.8s Lin Bytes test with Domain
[...]

PR #546, opened 2025-03-28
Lin bytes test fixes

### LINKED ISSUE #544: "Lin Bytes test with Thread" may fail because it is Thread unsafe
Up until #540 that improved the error finding ability of the Thread modes, we have considered "Lin Bytes test with Thread"
a positive test run with a low count of 250.

Now, the tooling has improved and the test may fail. It just did so on merging #542 to `main` running MSVC bytecode with `trunk`:
https://github.com/ocaml-multicore/multicoretests/actions/runs/14088717379/job/39459605635
```
random seed: 362773306

[ ]    0    0    0    0 / 5000     0.0s Lin Bytes test with Domain
[ ]    0    0    0    0 / 5000     0.0s Lin Bytes test with Domain (generating)
[ ]  509    0    0  509 / 5000   110.1s Lin Bytes test with Domain
[ ] 1112    0    0 1112 / 5000   170.6s Lin Bytes test with Domain
[✓] 1244    0    1 1243 / 5000   188.3s Lin Bytes test with Domain

[ ]    0    0    0    0 /  250     0.0s Lin Bytes test with Thread
[✗]   41    0    1   40 /  250    28.6s Lin Bytes test with Thread

[ ]    0    0    0    0 / 1000     0.0s Lin Bytes stress test with Domain
[ ]  931    0    0  931 / 1000    13.7s Lin Bytes stress test with Domain
[✓] 1000    0    0 1000 / 1000    14.7s Lin Bytes stress test with Domain

Negative test Lin Bytes test with Domain failed as expected (40 shrink steps):

[...]

--- Failure --------------------------------------------------------------------

Test Lin Bytes test with Thread failed (61 shrink steps):

                                  |           
                       Bytes.fill t 2 1 '\128'
                                  |           
                     .------------------------.
                     |                        |           
           Bytes.set t 0 '\128'    Bytes.is_valid_utf_8 t 
           Bytes.fill t 1 3 'U'                           


Messages for test Lin Bytes test with Thread:

  Results incompatible with sequential execution

                                        |                  
                        Bytes.fill t 2 1 '\128' : Ok (())  
                                        |                  
                     .------------------------------------.
                     |                                    |                  
      Bytes.set t 0 '\128' : Ok (())        Bytes.is_valid_utf_8 t : true    
      Bytes.fill t 1 3 'U' : Ok (())                                         

failure (1 tests failed, 0 tests errored, ran 3 tests)
File "src/bytes/dune", line 12, characters 7-16:
12 |  (name lin_tests)
            ^^^^^^^^^
(cd _build/default/src/bytes && .\lin_tests.exe --verbose)
Command exited with code 1.
```
I believe the problem here is that for `Bytes.is_valid_utf_8 t : true` a '\128' byte can only appear in valid utf-8
after a 110xxxxx 1110xxxx or 11110xxx prefix https://en.wikipedia.org/wiki/UTF-8#Description
This is not the case above, so I guess a Thread context switching situation is happening where
- Bytes.is_valid_utf_8 t is started in the second leg, that checks index 0 before `Bytes.set t 0 '\128'`
- Then context switches to the first leg, performing both the set and `Bytes.fill t 1 3 'U' : Ok (())` 
   effectively making this part of the string valid utf-8 and
- finally it context switches back to the second leg to confirm this in `Bytes.is_valid_utf_8 t`

... without the entire string never being valid utf-8 under any interleaving 😮 

Locally, by bumping up the `count` and running under a 5.3.0+bytecode switch I've been able to confirm this,
albeit it triggers rarely (I also commented out the other two Bytes test):
```
$ dune exec src/bytes/lin_tests.exe -- -v
random seed: 180290958
[✗]  955    0    1  954 / 5000     9.1s Lin Bytes test with Thread

--- Failure --------------------------------------------------------------------

Test Lin Bytes test with Thread failed (30 shrink steps):

                                  |           
                        Bytes.set t 2 '\128'  
                                  |           
                     .------------------------.
                     |                        |           
          Bytes.is_valid_utf_8 t    Bytes.set t 0 '\128'  
                                   Bytes.fill t 1 3 '\025'


Messages for test Lin Bytes test with Thread:

  Results incompatible with sequential execution

                                        |                  
                         Bytes.set t 2 '\128' : Ok (())    
                                        |                  
                     .------------------------------------.
                     |                                    |                  
       Bytes.is_valid_utf_8 t : true       Bytes.set t 0 '\128' : Ok (())    
                                          Bytes.fill t 1 3 '\025' : Ok (())  

failure (1 tests failed, 0 tests errored, ran 1 tests)
```

It is impressive that we are able to find such a corner-case for the `Bytes.is_valid_utf_8` predicate given that
- (a) we target 27 Bytes bindings
- (b) it involves a specific byte `'\128'` and
- (c) calls to both `is_valid_utf_8` and the right overwriting index in `Bytes.fill`

to observe a Thread-mode difference... 🙂 It also explains why this triggers somewhat rarely... 🤔

jmid:
The `Lin Bytes` tests are a pain point as witnessed by #544 #541 #526 #520 #518.

This PR improves on the situation by adding a `Bytes.to_seq` target which is a known source of domain-unsafety from the `STM.domain` test, yet was missing from the `Lin` tests. This addition means the `Lin_thread` test triggers much more reliably too, so it can be turned into an appropriate negative test.

Finally, shrink time of the `Lin Bytes Domain` test is a genuine issue #520 (it just happened again today). While trying to ramp up the targeted bindings of the corresponding STM Bytes tests today, I was playing with adding a `cmd` shrinker and started seeing similar longer shrink times when I added it for
```
 val Bytes.blit_string : string -> int -> t -> int -> int -> unit;
```

One problem is shrinking order: in STM, writing the shrinker by myself I could customize it to my liking, reducing the length and index integers first, and only the string secondly. Another problem is the `string` shrinker reducing each char repeatedly:
```
Shrink.string "zzzz" print_endline;;
zz
zz
zzz
nzzz
tzzz
wzzz
yzzz
znzz
ztzz
zwzz
zyzz
zznz
zztz
zzwz
zzyz
zzzn
zzzt
zzzw
zzzy
```
This will take quite a bit of time to reach the `'a'` goal for each entry, and repeatedly start over attempting to do so.
To improve on the situation, the PR therefore adds a quick-and-dirty char-shrinker for both strings and bytes in Lin, that simply attempts to reduce a non-`'a'` to `'a'`. As a result, shrinking for these should now work much faster out of the box.

Overall I'm hoping this fixes #544 #541 #520 #518 and thus represents a good step forward reliability-wise.
#526 is most likely due to Cygwin (signal?) issues and so remains open.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - #540: Significantly increase the probability of context switching in `Lin_thread`
         and `STM_thread` by utilizing `Gc.Memprof` callbacks. Avoid on 5.0-5.2
         without `Gc.Memprof` support.
+- #546: Speed up `Lin`'s default `string` and `bytes` shrinkers.
 
 ## 0.7
 
--- a/lib/lin.ml
+++ b/lib/lin.ml
@@ -159,6 +159,14 @@ let gen_deconstructible gen print eq = GenDeconstr (gen,print,eq)
 
 let qcheck_nat64_small = QCheck.(map Int64.of_int small_nat)
 
+(* QCheck's string shrinker reduces each char repeatedly which is too excessive for Lin *)
+let shrink_char c = QCheck.(if c = 'a' then Iter.empty else Iter.return 'a')
+let shrink_string = QCheck.Shrink.string ~shrink:shrink_char
+
+let string = QCheck.(set_shrink shrink_string string)
+let string_small = QCheck.(set_shrink shrink_string small_string)
+let string_small_printable = QCheck.(set_shrink shrink_string small_printable_string)
+
 let bytes_small_printable = QCheck.bytes_small_of QCheck.Gen.printable
 
 let unit =           GenDeconstr (QCheck.unit,           QCheck.Print.unit, (=))
@@ -174,9 +182,9 @@ let int32 =          GenDeconstr (QCheck.int32,          Int32.to_string,   Int3
 let int64 =          GenDeconstr (QCheck.int64,          Int64.to_string,   Int64.equal)
 let nat64_small =    GenDeconstr (qcheck_nat64_small,    Int64.to_string,   Int64.equal)
 let float =          GenDeconstr (QCheck.float,          QCheck.Print.float,  Float.equal)
-let string =         GenDeconstr (QCheck.string,         QCheck.Print.string, String.equal)
-let string_small =   GenDeconstr (QCheck.small_string,   QCheck.Print.string, String.equal)
-let string_small_printable = GenDeconstr (QCheck.small_printable_string, QCheck.Print.string, String.equal)
+let string =         GenDeconstr (string,                QCheck.Print.string, String.equal)
+let string_small =   GenDeconstr (string_small,          QCheck.Print.string, String.equal)
+let string_small_printable = GenDeconstr (string_small_printable, QCheck.Print.string, String.equal)
 let bytes =          GenDeconstr (QCheck.bytes,          QCheck.Print.bytes, Bytes.equal)
 let bytes_small =    GenDeconstr (QCheck.bytes_small,    QCheck.Print.bytes, Bytes.equal)
 let bytes_small_printable = GenDeconstr (bytes_small_printable, QCheck.Print.bytes, Bytes.equal)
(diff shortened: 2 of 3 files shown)


jmid:

jmid:

shym:
It seems that all the scheduled runs on 5.2 failed on this test this morning (I checked only 1/3 of the logs, I confess :-).
As 5.4 will enter freeze soon, maybe now is a reasonable time to retire 5.2 testing?

jmid:
My bad, sorry! :grimacing:  The new improved :tm: Thread mode utilizes `Gc.Memprof` support to trigger context switching. 5.3 restored `Memprof`, which means the thread test works well there and on `trunk` - both tested by CI on PRs.
I tested locally on 5.2 at some point and then forgot about it again. Given the above, I think the Lin Thread-mode test of `Bytes` should just be disabled on 5.2 and earlier.

PR #547, opened 2025-03-30
Expand on STM Bytes test

jmid:
This is the extension of the `STM Bytes` test mentioned in #546.
It targets 5 more bindings from the Bytes module - a decent increase over the existing 7.
I lost momentum when I hit the many index/contains/is_valid bindings covered by `src/bytes/lin_tests.ml`... :grimacing: 

~~While writing this, I noticed that `Util.Pp` only covers up to `cst3` - and similarly for `QCheck.Gen.map{,2,3}`.~~
~~I think it would be a good idea to extend both to cover up to `{cst,map}5`.~~ :slightly_smiling_face: Edit: fixed - see below

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -87,7 +87,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: c-cube/qcheck
-          ref: v0.23
+          ref: v0.25
           path: multicoretests/qcheck
 
       - name: Pre-Setup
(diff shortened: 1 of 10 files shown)


jmid:
> While writing this, I noticed that `Util.Pp` only covers up to `cst3` - and similarly for `QCheck.Gen.map{,2,3}`. I think it would be a good idea to extend both to cover up to `{cst,map}5`. 🙂

- I've now released qcheck-core.0.25 containing  `QCheck.Gen.map{,4,5}` and therefore bumped the lower bound to that, and removed the temporary `map4` definition again
- I've also added `Util.Pp.cst5` for uniformity and a CHANGES entry for that and `cst4`
- Finally I've rebased the PR on main, reordered and cleaned up the commits

jmid:

PR #538, opened 2025-03-07
CI: Pin dune in opam workflows too

jmid:
It has been ~3 weeks since #531, there hasn't been a dune release restoring `trunk` support yet, and I am now getting sufficiently annoyed having to repeatedly argue in CI failure runs.

This PR therefore pins dune to a trunk compatible version in the opam install workflows, akin to #531.

In the second commit I've also removed the "fail fast" mode, as in cases such as this, it is actually helpful to run all workflows to assess how many versions are affected as part of understanding the reason for a failure. I'd therefore hope to keep that commit, once we get a new `dune` release and remove #531 and 9d3625b526a9718558ba076fe4d51edb476028ac again.

PR #549, opened 2025-04-03
Remove dune pins again

### LINKED ISSUE #539: Remove dune pin in trunk workflows
`dune` support for `trunk` was broken causing us to pin a patched dune version
- for the `trunk` test workflows in #531 and
- for the `trunk` opam workflow in #538.

Once a `dune` release is made available including the patch, we should roll this pinning back.

jmid:
Fixes #539

`dune.3.18.0` is now available on the opam repo.

This rolls back the pinning needed in #531 and #538.

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -128,7 +128,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: ocaml/dune
-          ref: 2de53980efa731eacf5475139d9a5a218f9752a0
+          ref: 3.18.0
           path: dune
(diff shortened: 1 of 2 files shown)


jmid:
CI summary: all 36 workflows passed.

Side note: I spotted that Linux-ARM64 5.3 took a record low 11m (using caching) down from +16m :smiley: 
Looking at the log I then realized it used a cached entry and thus not dune.3.18.0 (recall: we did so to get CI running on trunk again). This is the first time we have had to pin dune though AFAIR, so IMO it seems overkill to make changes to the cache keys for this one-time case...

jmid:

Issue #283, opened 2023-01-16
[ocaml5-issue] BUS error on the Weak Hashset tests on macOS 5.0.0


jmid:
Ouch, `STM` tests trigger a BUS error on the `Weak` `Hashset` tests on macOS 5.0.0
https://github.com/ocaml-multicore/multicoretests/actions/runs/3913104092/jobs/6688588995
```
random seed: 470692489
generated error fail pass / total     time test name

[ ]    0    0    0    0 / 1000     0.0s STM Weak test sequential
[ ]    0    0    0    0 / 1000     0.0s STM Weak test sequential (generating)
[✓] 1000    0    0 1000 / 1000     0.3s STM Weak test sequential

[ ]    0    0    0    0 / 1000     0.0s STM Weak test parallel
[ ]   53    0    0   53 / 1000    63.2s STM Weak test parallel
[ ]  141    0    0  141 / 1000   124.4s STM Weak test parallel
[ ]  196    0    0  196 / 1000   186.8s STM Weak test parallel
[ ]  271    0    0  271 / 1000   252.0s STM Weak test parallel
[ ]  375    0    0  375 / 1000   315.7s STM Weak test parallel
[✓]  383    0    1  382 / 1000   336.4s STM Weak test parallel

[ ]    0    0    0    0 / 1000     0.0s STM Weak HashSet test sequential
[✓] 1000    0    0 1000 / 1000     0.4s STM Weak HashSet test sequential

[ ]    0    0    0    0 / 1000     0.0s STM Weak HashSet test parallel
[ ]   51    0    0   51 / 1000    38.9s STM Weak HashSet test parallel (shrinking:   19.0052)
[ ]   51    0    0   51 / 1000    98.9s STM Weak HashSet test parallel (shrinking:   23.0034)
[ ]   51    0    0   51 / 1000   159.0s STM Weak HashSet test parallel (shrinking:   25.0124)
[ ]   51    0    0   51 / 1000   219.1s STM Weak HashSet test parallel (shrinking:   27.0081)
[ ]   51    0    0   51 / 1000   279.1s STM Weak HashSet test parallel (shrinking:   29.0113)
[ ]   51    0    0   51 / 1000   339.2s STM Weak HashSet test parallel (shrinking:   32.0109)
[ ]   51    0    0   51 / 1000   399.3s STM Weak HashSet test parallel (shrinking:   35.0062)
[...]

PR #558, opened 2025-06-11
STM Weak test re-revision

jmid:
I recently had a fresh look at the weak array and weak hash set STM tests, and discovered a couple of things.
This PR is the outcome.

The weak array and weak hash set test originated in #214.
They've since been tweaked, e.g., in #365 which among other things switched to `int64` data for the weak hash sets test,
arguing that this was faster to shrink. As we no longer shrink the `data` this argument no longer holds.
Furthermore, the tests triggered crashes in the past with `string` `data` in #283, which was caused by a combination of race conditions and variable length memory records. This speaks for reverting and using `string`s for both tests. I've confirmed that both parallel tests can trigger segfaults on an older version (OCaml 5.0.0~alpha1) with these changes.

The PR also
- revises the post-conditions slightly,
- adjusts the frequencies slightly to skew them towards making additions, and
- cleans up the `protect` calls with thunks, rather than the current missing-last-argument-closure

### DIFF
--- a/src/weak/stm_tests_hashset_spec.ml
+++ b/src/weak/stm_tests_hashset_spec.ml
@@ -144,7 +144,7 @@ let postcond c (s:data list) res = match c, res with
   | Find d, Res ((Result (Int64,Exn),_),r) ->
     r = Error Not_found || (List.mem d s && r = Ok d)
   | Find_opt d, Res ((Result (Option Int64,Exn),_),r) ->
-    r = Ok None || r = Ok (Some d)
+    r = Ok None || (List.mem d s && r = Ok (Some d))
   | Find_all d, Res ((Result (List Int64,Exn),_),r) ->
     (match r with
      | Error _ -> false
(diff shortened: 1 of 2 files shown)


jmid:
CI summary for 7ce003b: all 13 trunk workflows (except trunk debug) failed due to #559
- 32bit trunk
- Bytecode trunk
- Cygwin trunk
- FP trunk
- Linux trunk
- Linux-ARM64 trunk
- macOS-ARM64 trunk
- macOS-intel trunk
- MinGW bytecode trunk
- MinGW trunk
- MSVC bytecode trunk
- MSVC trunk
- musl trunk

Out of 50 workflows, 13 failed with a genuine issue

jmid:
CI summary:
- Cygwin 5.3 timed out in `STM Dynarray stress test (int)`. I suspect this is due to the tearing observed in #528.

Out of 50 workflows 1 timed out with a supposedly genuine issue

PR #560, opened 2025-06-12
Rework GADT index types to get Lin compiling on trunk again

### LINKED ISSUE #559: [ocaml5-issue] GADT exhaustiveness check regression on trunk
Since 7 days ago `Lin` now fails to compile on `trunk` and instead errs with what appears to be a bogus non-exhaustiveness warning:
https://github.com/ocaml-multicore/multicoretests/actions/runs/15595919790/job/43926194200?pr=558
```
File "lib/lin.ml", lines 197-200, characters 2-81:
197 | ..match ty with
198 |   | Gen (arb, print) -> Gen (QCheck.option ?ratio arb, QCheck.Print.option print)
199 |   | GenDeconstr (arb, print, eq) -> GenDeconstr (QCheck.option ?ratio arb, QCheck.Print.option print, Option.equal eq)
200 |   | Deconstr (print, eq) -> Deconstr (QCheck.Print.option print, Option.equal eq)
Error (warning 8 [partial-match]): this pattern-matching is not exhaustive.
  Here is an example of a case that is not matched: State

[...]
```
This is preventing all `trunk` workflows from running ATM.
I first observed it on this past weekend's cron runs and now again on #558.

I've reduced it to the following reproducer and reported it upstream in https://github.com/ocaml/ocaml/issues/14081
```ocaml
type type1

type _ ty =
  | Cons1 : type1 ty
  | Cons2 : int ty

let f (ty : type1 ty) =
  match ty with
  | Cons1 -> Cons1
```

jmid:
Fixes #559 

ATM this changes two empty variant types in the interface, which is a bit annoying... :thinking:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Changes
 
+## NEXT RELEASE
+
+- #560: Change `Lin.{constructible,deconstructible}` from an empty variant type
+        to an abstract type to get `Lin` compiling on `5.5.0+trunk` again, due
+        to https://github.com/ocaml/ocaml/pull/13994 removing special handling
+        of abstract and empty variants defined in the current module.
+
 ## 0.8
 
 - #540: Significantly increase the probability of context switching in `Lin_thread`
--- a/lib/lin.ml
+++ b/lib/lin.ml
@@ -141,11 +141,11 @@ end
 
 (* Type-representing values *)
 
-type constructible = |
-type deconstructible = |
+type constructible = private Constructible [@@ocaml.warning "-37"]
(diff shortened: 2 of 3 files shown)


jmid:
@OlivierNicole Do you recall why you chose empty variant types for `constructible`/`deconstructible`
and abstract types for `combinable`/`noncombinable`?

Rather than rewrite the interface, I guess another way around this is to silence the warnings with attributes on each pattern match that `trunk` will now deem potentially non-exhaustive. We know from OCaml 5.0.0-5.4.0 that they are in fact exhaustive. On the positive side this will be less invasive, however it also smells like more of a temporary workaround... :shrug:

PR #561, opened 2025-06-19
GC test revision

jmid:
Analogous to #558 I also had another look at the `Gc` tests and spotted a couple of things there.
This PR is the result.

- The biggest change is presentational, where the generator pair is split and instead defined one after another.
- Second it replaces `minor_heap_size_gen` with a bit-shifting version rather than "choose from list literal".
- Third it generates `Counters` as part of `alloc_cmds` (like the other observational `Gc` functions) for consistency.

Finally there are a few smaller meaning preserving code clean ups.


jmid:

jmid:
CI summary for merge to `main`: all 51 workflows passed

PR #562, opened 2025-06-25
Add missing `l` and `L` suffixes in `int32` and `int64` printers for both `Lin` and `STM` 

jmid:
This PR adds missing `l` and `L` suffixes in `int32` and `int64` printers for both `Lin` and `STM`,
which meant that  `int32` and `int64` literals couldn't just be copy-pasted into, e.g., `utop`.

The fix simply forwards the calls to `QCheck.Print.int{32,64}`, thereby solving this in one central place.
These were furthermore some of the only remaining exceptions to type combinators not relying on `QCheck.Print`,
thus restoring OCD world order... :nerd_face: 

`QCheck.Print.int{32,64}` were added in 0.24 - but we already have 0.25 as a lower bound, so all should be well.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## NEXT RELEASE
 
+- #562: Fix the `int32` and `int64` printers in both `Lin` and `STM` to add
+        missing `l` and `L` suffixes on literals.
 - #560: Change `Lin.{constructible,deconstructible}` from an empty variant type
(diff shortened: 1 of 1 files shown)


jmid:
Note: The `Util` `int{32,64}` pp-functions used by most `STM` `show_cmd` functions already prints the suffixes.
Hence the patch has an impact mostly on `Lin` tests and `STM` combinator-annotated results.

PR #565, opened 2025-07-10
Bump QCheck to 0.26 (and fix float comparisons)

jmid:
The QCheck 0.26 release https://github.com/c-cube/qcheck/releases/tag/v0.26 changes the distribution of the `float` generator to avoid blind spots: c-cube/qcheck#350.

As we have a couple of tests with `float`s this PR bumps the QCheck version to see if we've missed anything `float`-related.

PR #566, opened 2025-07-28
Use `QCheck.Print` combinators in `Util.Pp` printers for consistency

jmid:
In #565 I discovered that `STM` argument and result printers differed causing a negative `nan` to display as `nan` as an argument to `Set`, yet it would print as `-nan` in the result, sending me on a goose chase. #562 was another example of outputs differing slightly.

Following the principle of least surprise, this PR switches all `Util.Pp` printers for base types to use the `QCheck.Print` combinators. I believe this further makes sense, since our opam packages are named with a `qcheck` prefix.

I don't claim that, e.g., the `QCheck.Print.float` printer is perfect as witnessed by the expect test output updates (e.g., `infinity` is valid copy-pasteble OCaml, whereas `inf` is not) but I would rather have to only improve such a combinator in a single, central location.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## NEXT RELEASE
 
+- #566: Use `QCheck.Print` combinators in `Util.Pp` too for consistency,
+        e.g., to avoid that `STM` argument and result printer outputs differ.
 - #562: Fix the `int32` and `int64` printers in both `Lin` and `STM` to add
         missing `l` and `L` suffixes on literals.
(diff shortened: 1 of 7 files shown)


jmid:
> I don't claim that, e.g., the `QCheck.Print.float` printer is perfect as witnessed by the expect test output updates (e.g., `infinity` is valid copy-pasteble OCaml, whereas `inf` is not) but I would rather have to only improve such a combinator in a single, central location.

I've opened an upstream issue on QCheck here: https://github.com/c-cube/qcheck/issues/353

jmid:
CI summary for ddcebf1: All 50 workflows passed

PR #567, opened 2025-08-11
Adjust Gc.stat stack_size check after 5.5 restoration

jmid:
The test suite is failing on 5.5+trunk on the `Gc` tests.
This is due to the `stack_size` entry of `Gc.stat` now being restored in https://github.com/ocaml/ocaml/pull/14168 which would previously be constantly zero on OCaml 5.
The PR relaxes the check for the entry for 5.5 and upward.


jmid:

